PR #27, opened 2024-03-29
Initial pass at docs before going public

denelon:
This is some of the initial work to make sure all the required documents are in place prior to making the repository public. I've added some base templates as well.


moonickan-sh (commented):
https://github.com/microsoft/winget-dsc/pull/27/commits/3ea757976d8aef6062960af049ab0ff7b8e42dd8#diff-eca12c0a30e25b4b46522ebf89465a03ba72a03f540796c979137931d8f92055

PR #42, opened 2024-06-17
Run Pester tests in build pipeline

ryfu-msft:
Adds tasks to run and publish pester unit tests.


florelis (approved):
> Fixes the pipeline so that governance checks only run once per build not for each artifact that is published.

What change does that?

ryfu-msft:
> > Fixes the pipeline so that governance checks only run once per build not for each artifact that is published.
> 
> What change does that?

Old comment, updated the description to remove that.

PR #57, opened 2024-09-10
Microsoft.VSCode.DSC initial implementation

ryfu-msft:
Adds a new DSC resource for VSCode that utilizes the code.exe executable to install a specified VSCode extension. 

Accepts 2 parameters
- Name of the extension (required)
- Version of the extension (optional)

Tested manually with various VSCode extensions.

### DIFF
--- a/resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
+++ b/resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
@@ -4,7 +4,7 @@
 $ErrorActionPreference = "Stop"
 Set-StrictMode -Version Latest
 
-enum Ensure
+enum VSCodeEnsure
 {
     Absent
     Present
@@ -21,45 +21,34 @@ class VSCodeExtension
     [string] $Version
 
     [DscProperty()]
-    [Ensure] $Ensure = [Ensure]::Present
-
-    [DscProperty(NotConfigurable)]
-    [hashtable] $InstalledExtensions = @{}
+    [VSCodeEnsure] $Ensure = [VSCodeEnsure]::Present
 
     [VSCodeExtension] Get()
     {
-        Assert-VSCode
-
         $currentState = [VSCodeExtension]::new()
-        $currentState.Ensure = [Ensure]::Absent
+        $currentState.Ensure = [VSCodeEnsure]::Absent
         $currentState.Name = $this.Name
         $currentState.Version = $this.Version
+
+        $installedExtensions = @{}
         $extensionList = (Invoke-VSCode -Command "--list-extensions --show-versions") -Split [Environment]::NewLine
 
+        # Populate hash table with installed VSCode extensions.
         foreach ($extension in $extensionList)
         {
             $info = $extension -Split '@'
             $extensionName = $info[0]
             $extensionVersion = $info[1]
-            $currentState.InstalledExtensions[$extensionName] = $extensionVersion
+            $installedExtensions[$extensionName] = $extensionVersion
         }
 
-        foreach ($extension in $currentState.InstalledExtensions.Keys)
+        foreach ($extension in $installedExtensions.Keys)
         {
             if ($extension -eq $this.Name)
             {
-                if ($null -ne $this.Version)
-                {
-                    if ($this.Version -eq $currentState.InstalledExtensions[$this.Name])
-                    {
-                        $currentState.Ensure = [Ensure]::Present
-                    }
-                }
-                else
-                {
-                    $currentState.Ensure = [Ensure]::Present
-                }
-
+                $currentState.Ensure = [VSCodeEnsure]::Present
+                $currentState.Name = $extension
+                $currentState.Version = $installedExtensions[$extension]
                 break
             }
         }
@@ -70,7 +59,17 @@ class VSCodeExtension
     [bool] Test()
     {
         $currentState = $this.Get()
-        return $currentState.Ensure -eq $this.Ensure
+        if ($currentState.Ensure -ne $this.Ensure)
+        {
+            return $false
+        }
+
+        if ($null -ne $this.Version -and $this.Version -ne $currentState.Version)
+        {
+            return $false
+        }
+
+        return $true
     }
 
     [void] Set()
@@ -80,7 +79,7 @@ class VSCodeExtension
             return
         }
 
-        if ($this.Ensure -eq [Ensure]::Present)
+        if ($this.Ensure -eq [VSCodeEnsure]::Present)
         {
             $extensionArg = $this.Name
 
@@ -95,22 +94,6 @@ class VSCodeExtension
         {
             Invoke-VSCode -Command "--uninstall-extension $($this.Name)"
         }
-
-    }
-}
-
-function Assert-VSCode
-{
-    # Refresh session $PATH value before invoking 'code.exe'.
-    $env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine") + ";" + [System.Environment]::GetEnvironmentVariable("Path","User")
-    try
-    {
-        Invoke-VSCode -Command '--help'
-        return
-    }
-    catch
-    {
-        throw "VSCode is not installed"
     }
 }
 
@@ -121,5 +104,12 @@ function Invoke-VSCode
         [string]$Command
     )
 
-    return Invoke-Expression -Command "code $Command"
+    try 
+    {
+        return Invoke-Expression "& `"$env:LocalAppData\Programs\Microsoft VS Code\bin\code.cmd`" $Command"
+    }
+    catch
+    {
+        throw "VSCode is not installed."
+    }
 }
--- a/resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
+++ b/resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
@@ -14,7 +14,6 @@ enum Ensure
 [DSCResource()]
 class VSCodeExtension
 {
-    # Key required. Do not set.
     [DscProperty(Key)]
     [string] $Name
 
@@ -29,9 +28,6 @@ class VSCodeExtension
 
     [VSCodeExtension] Get()
     {
-        # Get all installed extensions
-        # Check if visual studio code is installed.
-
         Assert-VSCode
 
         $currentState = [VSCodeExtension]::new()
@@ -52,7 +48,6 @@ class VSCodeExtension
         {
             if ($extension -eq $this.Name)
             {
-                # if version parameter is provided.
                 if ($null -ne $this.Version)
                 {
                     if ($this.Version -eq $currentState.InstalledExtensions[$this.Name])
@@ -106,7 +101,7 @@ class VSCodeExtension
 
 function Assert-VSCode
 {
-    # Refresh session $path value before invoking 'code'
+    # Refresh session $PATH value before invoking 'code.exe'.
     $env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine") + ";" + [System.Environment]::GetEnvironmentVariable("Path","User")
     try
     {
--- a/resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
+++ b/resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
@@ -4,11 +4,101 @@
 $ErrorActionPreference = "Stop"
 Set-StrictMode -Version Latest
 
+#region Enums
 enum VSCodeEnsure
 {
     Absent
     Present
 }
+#endregion Enums
+
+#region Functions
+function Get-VSCodeCLIPath
+{
+    $codeCLIUserPath = "$env:LocalAppData\Programs\Microsoft VS Code\bin\code.cmd"
+    $codeCLIMachinePath = "$env:ProgramFiles\Microsoft VS Code\bin\code.cmd"
+
+    if (Test-Path -Path $codeCLIUserPath)
+    {
+        return $codeCLIUserPath
+    }
+    elseif (Test-Path -Path $codeCLIMachinePath)
+    {
+        return $codeCLIMachinePath
+    }
+    else
+    {
+        throw "VSCode is not installed."
+    }
+}
+
+function Install-VSCodeExtension
+{
+    [CmdletBinding()]
+    param (
+        [Parameter(Mandatory, ValueFromPipelineByPropertyName)]
+        [string]$Name,
+        [Parameter(ValueFromPipelineByPropertyName)]
+        [string]$Version
+    )
+    
+    begin
+    {
+        function Get-VSCodeExtensionInstallArgument
+        {
+            param([string]$Name, [string]$Version)
+            
+            if ([string]::IsNullOrEmpty($Version))
+            {
+                return $Name
+            }
+
+            return @(
+                $Name
+                $Version
+            ) -join '@'
+        }
+    }
+    
+    process
+    {
+        $installArgument = Get-VSCodeExtensionInstallArgument @PSBoundParameters
+        Invoke-VSCode -Command "--install-extension $installArgument"
+    }
+}
+
+function Uninstall-VSCodeExtension
+{
+    [CmdletBinding()]
+    param (
+        [Parameter(Mandatory, ValueFromPipelineByPropertyName)]
+        [string]$Name
+    )
+        
+    Invoke-VSCode -Command "--uninstall-extension $($this.Name)"  
+}
+
+function Invoke-VSCode
+{
+    param (
+        [Parameter(Mandatory = $true)]
+        [string]$Command
+    )
+
+    try 
+    {
+        Invoke-Expression "& `"$VSCodeCLIPath`" $Command"
+        "$env:ProgramFiles\Microsoft VS Code\bin\code.cmd"
+    }
+    catch
+    {
+        throw "Executing code.exe with {$Command} failed."
+    }
+}
+#endregion Functions
+
+# Keeps the path of the code.exe CLI path.
+$VSCodeCLIPath = Get-VSCodeCLIPath
 
 #region DSCResources
 [DSCResource()]
@@ -106,19 +196,6 @@ class VSCodeExtension
         }          
     }
 
-    [string] GetInstallArgument()
-    {
-        if ($null -eq $this.Version)
-        {
-            return $this.Name
-        }
-
-        return @(
-            $this.Name
-            $this.Version
-        ) -join '@'
-    }
-
     [void] Install([bool] $preTest)
     {
         if ($preTest -and $this.Test())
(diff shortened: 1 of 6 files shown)


resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,125 @@
...
+    [string] $Name
+
+    [DscProperty()]
+    [string] $Version
+
+    [DscProperty()]
+    [Ensure] $Ensure = [Ensure]::Present
+
+    [DscProperty(NotConfigurable)]
+    [hashtable] $InstalledExtensions = @{}
```
msftrubengu, line 27: Why do you need this as a member? You are only in Get and it seems that you just iterate over then to find it.

ryfu-msft: I wanted to have a way to show the users what extensions are currently installed along with their versions. I made it NotConfigurable so they can't modify it and it is only to be used as a reference. I figured more info is better than none even if it is not absolutely necessary.

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,125 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+
+$ErrorActionPreference = "Stop"
+Set-StrictMode -Version Latest
+
+enum Ensure
```
JohnMcPMS, line 7: Enums and classes are not namespaced when loaded into PowerShell; you should use a more specific name like `VSCodeEnsure`.

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
JohnMcPMS: This is export functionality, which we don't have right now but may in the future.  Remove the property.

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,125 @@
...
+        {
+            $info = $extension -Split '@'
+            $extensionName = $info[0]
+            $extensionVersion = $info[1]
+            $currentState.InstalledExtensions[$extensionName] = $extensionVersion
+        }
+
+        foreach ($extension in $currentState.InstalledExtensions.Keys)
+        {
+            if ($extension -eq $this.Name)
```
JohnMcPMS, line 49: Should this be a case-sensitive comparison?  That should depend on whether code.exe treats it as case-sensitive.

May not matter if there is a way to pass `Name` to code.exe to get just the one we care about though.

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,125 @@
...
+            {
+                if ($null -ne $this.Version)
```
JohnMcPMS, line 51: `Version` is not a key and therefore does not drive the Present/Absent decision for `Get`.  `Get` uses only keys to find the resource, then returns the other properties set with the values that would best reflect what should be passed to `Test` to return `$true` or `Set` to produce the current state.

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,125 @@
...
+    [DscProperty(NotConfigurable)]
+    [hashtable] $InstalledExtensions = @{}
+
+    [VSCodeExtension] Get()
+    {
+        Assert-VSCode
+
+        $currentState = [VSCodeExtension]::new()
+        $currentState.Ensure = [Ensure]::Absent
+        $currentState.Name = $this.Name
```
JohnMcPMS, line 35: I preferred using the value returned by the underlying data store when it was found, mostly for casing.

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,125 @@
...
+            }
+        }
+        
+        return $currentState
+    }
+
+    [bool] Test()
+    {
+        $currentState = $this.Get()
+        return $currentState.Ensure -eq $this.Ensure
```
JohnMcPMS, line 73: Per previous comments, needs to check `Version` as well as `Ensure`.

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,125 @@
...
+    }
+
+    [void] Set()
+    {
+        if ($this.Test())
```
JohnMcPMS, line 78: Depending on how code.exe handles it, you should consider just always calling it.

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,125 @@
...
+            Invoke-VSCode -Command "--uninstall-extension $($this.Name)"
+        }
+
+    }
+}
+
+function Assert-VSCode
+{
+    # Refresh session $PATH value before invoking 'code.exe'.
+    $env:Path = [System.Environment]::GetEnvironmentVariable("Path","Machine") + ";" + [System.Environment]::GetEnvironmentVariable("Path","User")
```
JohnMcPMS, line 105: This seems like it could cause issues.  What if I updated `%PATH%` before invoking this DSC and now you've lost my change that pointed you to code.exe?

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,125 @@
...
+        }
+        else
+        {
+            Invoke-VSCode -Command "--uninstall-extension $($this.Name)"
+        }
+
+    }
+}
+
+function Assert-VSCode
```
JohnMcPMS, line 102: Why not just have `Invoke-VSCode` handle everything that this does?  Assert is using Invoke anyway.

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
ryfu-msft: Removed property

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
ryfu-msft: Changed to `VSCodeEnsure`

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
ryfu-msft: code.exe treats it as case-insensitive, so I left this as is

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
ryfu-msft: Updated so that Version does not affect the Present/Absent state. Version match is now checked in the Test() method

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
ryfu-msft: Value is now updated with underlying data store if extension is found.

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
ryfu-msft: I did this initially because the path was not getting updated if the previous configuration unit was install VSCode. I changed it so that it points directly to the expected code.cmd path which is this `$env:LocalAppData\Programs\Microsoft VS Code\bin\code.cmd` to avoid messing with the path values.

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
ryfu-msft: Removed Assert-VSCode and applied checks to Invoke-VSCode

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
ryfu-msft: Not sure if I follow, but Test calls Get, which always calls code.exe. I'm probably misunderstanding something though...

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,115 @@
...
+
+    try 
+    {
+        return Invoke-Expression "& `"$env:LocalAppData\Programs\Microsoft VS Code\bin\code.cmd`" $Command"
+    }
+    catch
+    {
+        throw "VSCode is not installed."
+    }
+}
```
michaeltlombardi, line 115: A few questions about this utility function:

- What if `code` is installed, but not in this location?
- What happens if the command is given invalid flags?
- What happens if the command is given valid flags but fails for another reason (e.g. can't connect to the internet)?
- What output does this return? Can that output be meaningfully parsed?
- Is this utility function intended for use for other resources managing VS Code, or just this one? Can this function shift any of the logic for defining arguments into the function?
- If you find it easier to test functions than classes in Pester, you could extract the logic for getting all extensions, installing an extension, and uninstalling an extension into functions that accept the name/version parameters and pass the instance properties or pipe the instance to the function, i.e.:

  ```powershell
  function Install-VSCodeExtension
  {
      [CmdletBinding()]
      param (
        [Parameter(Mandatory, ValueFromPipelineByPropertyName)]
        [string]$Name,
        [Parameter(ValueFromPipelineByPropertyName)]
        [string]$Version
      )
      
      # Implementation
  }
  
  $extension = [VSCodeExtension]@{
      Name    = 'Foo'
      Version = '1.2.3'
  }
  
  # Direct pass:
  Install-VSCodeExtension -Name $extension.Name -Version $extension.Version
  
  # Pipeline:
  $extension | Install-VSCodeExtension
  ```

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,115 @@
...
+
+enum VSCodeEnsure
+{
+    Absent
+    Present
+}
+
+#region DSCResources
+[DSCResource()]
+class VSCodeExtension
```
michaeltlombardi, line 15: Recommend adding a static property and method for determining whether VS Code is installed/available for testing and simplified maintenance - the property should just hold the value for the check, and have it set in the static constructor.

michaeltlombardi (commented):
Added a few notes on possible areas for extraction/extension/performance/testability, but overall this seems to be a very functional implementation.

My largest concern is around surfacing useful error messages when an install operation fails - in particular, the resource should be able to indicate:

- That the install failed because the machine can't connect to the internet.
- That the install failed because the specified extension name doesn't exist.
- That the install failed because the specified extension version doesn't exist.
- That the install failed for another reason, and surface the output from VS Code's command (and exit code).

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,115 @@
...
+            {
+                $extensionArg += "@$($this.Version)"
+            }
+
+            Invoke-VSCode -Command "--install-extension $($extensionArg)"
+        }
+        else
+        {
+            Invoke-VSCode -Command "--uninstall-extension $($this.Name)"
+        }
```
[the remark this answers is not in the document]
ryfu-msft: Implemented the suggested changes. Thank you for the very thorough code snippet :)

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,115 @@
...
+        foreach ($extension in $installedExtensions.Keys)
+        {
+            if ($extension -eq $this.Name)
+            {
+                $currentState.Ensure = [VSCodeEnsure]::Present
+                $currentState.Name = $extension
+                $currentState.Version = $installedExtensions[$extension]
+                break
+            }
+        }
```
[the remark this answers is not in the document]
ryfu-msft: Changed to suggested

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,115 @@
...
+
+    [VSCodeExtension] Get()
+    {
+        $currentState = [VSCodeExtension]::new()
+        $currentState.Ensure = [VSCodeEnsure]::Absent
+        $currentState.Name = $this.Name
+        $currentState.Version = $this.Version
+
+        $installedExtensions = @{}
+        $extensionList = (Invoke-VSCode -Command "--list-extensions --show-versions") -Split [Environment]::NewLine
```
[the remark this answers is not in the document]
ryfu-msft: Added the static export method as suggested.

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,191 @@
...
+    [CmdletBinding()]
+    param (
+        [Parameter(Mandatory, ValueFromPipelineByPropertyName)]
+        [string]$Name,
+        [Parameter(ValueFromPipelineByPropertyName)]
+        [string]$Version
+    )
+        
+    Invoke-VSCode -Command "--install-extension $($this.GetInstallArgument())"
+}
```
michaeltlombardi, line 163: The `$this` variable in the function's scope will be `$null`, which I realize I must've missed when I was advising extraction - You could create a `Get-VSCodeExtensionInstallArgument` function to hold that logic in this function's `begin` block, and pass the `$PSBoundParameters` to it:

```suggestion
function Install-VSCodeExtension
{
    [CmdletBinding()]
    param (
        [Parameter(Mandatory, ValueFromPipelineByPropertyName)]
        [string]$Name,
        [Parameter(ValueFromPipelineByPropertyName)]
        [string]$Version
    )
    
    begin
    {
        function Get-InstallArgument
        {
            param([string]$Name, [string]$Version)

            if ([string]::IsNullOrEmpty($this.Version))
            {
                return $Name
            }

            return @(
                $Name
                $Version
            ) -join '@'
        }
    }
    
    process
    {
        $installArgument = Get-VSCodeExtensionInstallArgument @PSBoundParameters

        Invoke-VSCode -Command "--install-extension $installArgument"
    }
}
```

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,191 @@
...
+
+function Uninstall-VSCodeExtension
+{
+    [CmdletBinding()]
+    param (
+        [Parameter(Mandatory, ValueFromPipelineByPropertyName)]
+        [string]$Name
+    )
+        
+    Invoke-VSCode -Command "--uninstall-extension $($this.Name)"
```
michaeltlombardi, line 173:
```suggestion
    Invoke-VSCode -Command "--uninstall-extension $Name"  
```

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
ryfu-msft: I created a new function called `Get-VSCodeCLIPath` that does two things,:
1. check the existence of VSCode installation
2. returns the path of the code.cmd file. 

I call this at the beginning of the module file and assign it to a local variable to by used by `Invoke-VSCode`. The reason I did it this way is so that we only have to determine the code.exe path once and all the functions that interact with the cli could be utilized by future DSC resources. I felt that a static variable was not entirely necessary because almost everything is driven though the interaction with the code.exe cli so checking for its existence should be a requirement.

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
ryfu-msft: Code can only be installed in two locations based on scope. 
`C:\Users\<Your Username>\AppData\Local\Programs\Microsoft VS Code`
`C:\Program Files\Microsoft VS Code`

Code.exe doesn't return a proper error message if you provide invalid arguments or even an invalid extension name so it is difficult to produce a reasonable error message from the output. I agree that this is another concern that I have. 

The utility functions that we have here are intended to be used for other DSC resources for VSCode. This prototype that I have here is just for adding VSCode extensions, but the code.exe cli tool is very powerful so I'm sure there are more things we can add to this DSC module.

Changed to suggested functions for easier testing.

michaeltlombardi: Those are the two locations for VS Code _on Windows when installed with WinGet_, right? But the logic for installing/updating VS Code extensions is the same when a user installs VS Code in [portable mode] or on macOS/Linux - I'm not too worried about this, but it's worth keeping in mind that aside from this code snippet, everything else I see in here works fully across platforms and install methodologies, and PSDSC resources published to the gallery, especially for managing something like VS Code, are likely to draw users this code won't support.

I think both deferring this problem and/or just saying "Windows only" support for the module and resources are okay, to be clear.

[portable mode]: https://code.visualstudio.com/docs/editor/portable

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
```diff
@@ -0,0 +1,227 @@
...
+{
+    param (
+        [Parameter(Mandatory = $true)]
+        [string]$Command
+    )
+
+    try 
+    {
+        Invoke-Expression "& `"$VSCodeCLIPath`" $Command"
+        "$env:ProgramFiles\Microsoft VS Code\bin\code.cmd"
```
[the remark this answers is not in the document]
ryfu-msft: Removed. That was a line leftover from debugging 😓

resources/Microsoft.VSCode.DSC/Microsoft.VSCode.DSC.psm1
ryfu-msft: Good callout. I left a todo comment on the function stating that this is only supported on user/machine installs for 'Windows'. I'll need to figure out a better way to determine the install location especially to support the portable install scenario.

ryfu-msft:
Will address tests in a separate PR as this is just a prototype.

Issue #59, opened 2024-09-18
Support VSCode Insiders - Microsoft.VSCode.Dsc


Gijsreyn:
### Description of the new feature / enhancement

With the new `Microsoft.VSCode.Dsc` released 4 days ago, I was curious if VSCode Insiders will also be able to be configured using this resource.

denelon:
@Gijsreyn thanks for pointing this out. We'll add support for the insider release.

Gijsreyn:
@denelon Added a pull request. I had fiddled around with adding some sort of switch on the `VSCodeExtension` class, but noticed my class-based programming as sharp is I would have thought.

Anyway, having it as seperate class, always keeps a clear distinction between the two. I have moved away the function call to determine the path. It's called now when the methods are invoked. I included some Pester test based on the references from `Microsoft.Windows.Developer` module.

If I have some time left, I might also include the part for macOS/Linux.

Gijsreyn:
@denelon I think this one can be closed. I hit up the [PR](https://github.com/microsoft/winget-dsc/pull/60) and it is merged.

PR #60, opened 2024-09-27
Added new VSCodeInsiders class with Pester tests

Gijsreyn:

-----

Addresses issue #59

### DIFF
--- a/resources/Microsoft.VSCode.Dsc/Microsoft.VSCode.Dsc.psm1
+++ b/resources/Microsoft.VSCode.Dsc/Microsoft.VSCode.Dsc.psm1
@@ -118,7 +118,6 @@ class VSCodeExtension {
         $this.Version = $Version
     }
 
-    # TODO: validate if 'dsc.exe' is able to parse in the 'VSCodeExtension' class with UseInsiders input.
     [VSCodeExtension[]] Export([bool]$UseInsiders)
     {
         if ($UseInsiders) {
@@ -199,7 +198,7 @@ class VSCodeExtension {
         }
 
         Install-VSCodeExtension -Name $this.Name -Version $this.Version
-        [VSCodeExtension]::GetInstalledExtensions()
+        [VSCodeExtension]::GetInstalledExtensions($this.UseInsiders)
     }
 
     [void] Install() {
@@ -208,7 +207,7 @@ class VSCodeExtension {
 
     [void] Uninstall([bool] $preTest) {
         Uninstall-VSCodeExtension -Name $this.Name
-        [VSCodeExtension]::GetInstalledExtensions()
+        [VSCodeExtension]::GetInstalledExtensions($this.UseInsiders)
     }
 
     [void] Uninstall() {
--- a/resources/Microsoft.VSCode.Dsc/Microsoft.VSCode.Dsc.psm1
+++ b/resources/Microsoft.VSCode.Dsc/Microsoft.VSCode.Dsc.psm1
@@ -105,8 +105,7 @@ class VSCodeExtension {
     [bool] $Exist = $true
 
     [DscProperty()]
-    [bool]
-    $UseInsiders = $false
+    [bool] $UseInsiders = $false
 
     static [hashtable] $InstalledExtensions
 
--- a/resources/Microsoft.VSCode.Dsc/Microsoft.VSCode.Dsc.psm1
+++ b/resources/Microsoft.VSCode.Dsc/Microsoft.VSCode.Dsc.psm1
@@ -7,21 +7,11 @@ Set-StrictMode -Version Latest
 #region Functions
 function Get-VSCodeCLIPath {
     param (
-        [switch]$UseInsiders,
-        [string]$CustomPath
+        [switch]$UseInsiders
     )
 
     # Currently only supports user/machine install for VSCode on Windows.
     # TODO: Update this function to handle when VSCode is installed in portable mode or on macOS/Linux.
-
-    # If a custom path is provided, use it
-    if ($PSBoundParameters.ContainsKey('CustomPath')) { 
-        if ((Test-Path -Path $CustomPath) -and (Split-Path $CustomPath -Leaf) -in ('code.cmd', 'code-insiders.cmd')) {
-            return $CustomPath
-        } else {
-            throw "The specified custom path does not exist or does not start with 'code.cmd' or 'code-insiders.cmd'."
-        }
-    }
 
     # Determine the paths based on whether the Insiders version is used
     if ($UseInsiders) {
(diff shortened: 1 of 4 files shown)


resources/Microsoft.VSCode.Dsc/Microsoft.VSCode.Dsc.psm1
```diff
@@ -90,9 +123,6 @@ function Invoke-VSCode
 }
 #endregion Functions
 
-# Keeps the path of the code.exe CLI path.
-$VSCodeCLIPath = Get-VSCodeCLIPath
-
```
ryfu-msft, line 95: I think it might be better to add a boolean parameter like `$UseInsiders` instead of creating a completely new DSC resource. That way we don't have to maintain two resources that are essentially doing the same thing.

You can use the boolean parameter inside the static constructor to check if `$UseInsiders = $true` and set $VSCodeCLIPath with the correct insiders code.exe path. So basically:

```powershell
if $UseInsiders
{
    $VSCodeCLIPath = Get-VSCodeInsidersCLIPath
}
```

ryfu-msft (changes requested):

resources/Microsoft.VSCode.Dsc/Microsoft.VSCode.Dsc.psm1
Gijsreyn: @ryfu-msft Thanks for the review. This is exactly what I wanted to do in the first place, but couldn't grasp my head around it. My class-based PowerShell development is a bit frisky.

Wouldn't it require to add an additional property in the class, allowing the ability to choose between Insiders or Stable?

ryfu-msft: Yup, you would need to add another boolean DSCProperty and use that to determine whether we point to the Insiders or Stable VSCodeCLI path.

Gijsreyn: @ryfu-msft  Hey Ryan, I've rewritten the `Get-VSCodeCLIPath` (also started on the custom path one from #61 ). I'm struggling with how to get it initialized. Whenever `Initialize` is called, a loop occurs. I'm seeing something over the head. Do you mind taking a look?

Gijsreyn: After struggling for some time now, I reconstructed the constructors. Because of the following logic when calling Invoke-DscResource:

image

The call is made later to determine the CLI path. The export has an overload added. It's good to check if this is possible with `dsc.exe` I guess.

resources/Microsoft.VSCode.Dsc/Microsoft.VSCode.Dsc.psm1
```diff
@@ -106,115 +114,124 @@ class VSCodeExtension
...
+    [System.String]
+    $CustomPath
+
     static [hashtable] $InstalledExtensions
 
-    static VSCodeExtension()
-    {
-        [VSCodeExtension]::GetInstalledExtensions()
+    VSCodeExtension() {
     }
```
ryfu-msft, line 130: I'm okay with moving GetInstalledExtensions() to the get method. Since we are no longer using a static constructor, this can be removed.

resources/Microsoft.VSCode.Dsc/Microsoft.VSCode.Dsc.psm1
```diff
@@ -106,115 +114,124 @@ class VSCodeExtension
...
         [VSCodeExtension]::GetInstalledExtensions()
     }
 
-    [void] Uninstall()
-    {
+    [void] Uninstall() {
         $this.Uninstall($true)
     }
-#endregion VSCodeExtension helper functions
+    #endregion VSCodeExtension helper functions
```
ryfu-msft, line 235:
```suggestion
#endregion VSCodeExtension helper functions
```

resources/Microsoft.VSCode.Dsc/Microsoft.VSCode.Dsc.psm1
```diff
@@ -106,115 +114,124 @@ class VSCodeExtension
...
     }
 
-    static [VSCodeExtension[]] Export()
+    VSCodeExtension([string]$Name, [string]$Version) {
+        $this.Name = $Name
+        $this.Version = $Version
+    }
+
+    # TODO: validate if 'dsc.exe' is able to parse in the 'VSCodeExtension' class with UseInsiders input.
+    static [VSCodeExtension[]] Export([System.Boolean]$UseInsiders, [System.String]$CustomPath)
```
ryfu-msft, line 138: This method no longer needs to be static anymore since you're now calling it inside the get method.
```suggestion
    [VSCodeExtension[]] Export([bool]$UseInsiders)
```

resources/Microsoft.VSCode.Dsc/Microsoft.VSCode.Dsc.psm1
```diff
@@ -106,115 +114,124 @@ class VSCodeExtension
...
-    {
-        [VSCodeExtension]::GetInstalledExtensions()
+    VSCodeExtension() {
     }
 
-    static [VSCodeExtension[]] Export()
+    VSCodeExtension([string]$Name, [string]$Version) {
+        $this.Name = $Name
+        $this.Version = $Version
+    }
```
ryfu-msft, line 135: Remove as this is covered by the default constructor.

tests/Microsoft.VSCode.Dsc/Microsoft.Windows.Dsc.Tests.ps1
```diff
@@ -0,0 +1,138 @@
...
+    Invoke-WebRequest https://raw.githubusercontent.com/PowerShell/vscode-powershell/main/scripts/Install-VSCode.ps1 -UseBasicParsing -OutFile Install-VSCode.ps1
+    .\Install-VSCode.ps1 -BuildEdition Stable-User -Confirm:$false
+
+    # Install VSCode Insiders
+    .\Install-VSCode.ps1 -BuildEdition Insider-User -Confirm:$false
+}
+
+Describe 'List available DSC resources' {
+    It 'Shows DSC Resources' {
+        $expectedDSCResources = "VSCodeExtension", "VSCodeInsidersExtension"
```
ryfu-msft, line 27:
```suggestion
        $expectedDSCResources = "VSCodeExtension"
```

ryfu-msft (changes requested):

resources/Microsoft.VSCode.Dsc/Microsoft.VSCode.Dsc.psm1
Gijsreyn: @ryfu-msft  I didn't totally get what you mean what is required to be removed.

Gijsreyn:
Can you run the pipeline once more?

ryfu-msft (changes requested):
These should fix your issues with the test. Be sure to actually run the tests on your machine before submitting.

Gijsreyn:
> These should fix your issues with the test. Be sure to actually run the tests on your machine before submitting.

Sorry Ryan, was struggling with the Pester version and getting it running. Should be fixed now. Can you run it once more?

ryfu-msft (approved):
Looks great! Huge thank you for this contribution 😄 and for the tests.

Gijsreyn:
And you thanks for the checks. Looking forward to seeing the custom path coming :)

PR #87, opened 2024-11-01
Add some tests for GitDsc

Trenly:

I noticed that the `GitConfigFile` resource in the `GitDsc` module didn't actually exist. So, I removed it and added the test for listing the available resources. I also added a basic test for the `GitClone` resource. I wasn't familiar enough with the other resources in the module to add tests for them, but I figured that some tests were better than none. If this PR merges, I'll raise an issue for tracking that the additional tests should be added

-----


Trenly:
o.O They ran locally for me. . . Time to debug!

ryfu-msft:
> o.O They ran locally for me. . . Time to debug!

You might need to rearrange the order so that the helper functions are first before the Assert-Git

Trenly:
I think the issue is actually that Git just isn't installed, so there is no way for it to actually clone the repository.
I think this leaves 2 options - 
1) Have the tests install Git using Microsoft.WinGet.DSC
2) Have some test data or test files that can be copied to the directory in place of asserting that the actual process of running `git clone` works

Trenly:
Okay, that was harder than I thought it would be, since Pester doesn't like to play nice if you don't have the scopes set juuuuuust right, but I think I got it to work

resources/GitDsc/GitDsc.psd1
```diff
@@ -8,130 +8,129 @@
...
 
-# Script module or binary module file associated with this manifest.
-RootModule = 'GitDsc.psm1'
+    # Script module or binary module file associated with this manifest.
+    RootModule           = 'GitDsc.psm1'
 
-# Version number of this module.
-ModuleVersion = '0.1.0'
+    # Version number of this module.
+    ModuleVersion        = '0.1.1'
```
ryfu-msft, line 15: I just caught this but don't increment the version on any of these manifest files. We have an internal publishing pipeline that will do that for us.

ryfu-msft (changes requested):
Just one fix and the rest looks good

resources/GitDsc/GitDsc.psd1
Trenly: Sounds good!

Trenly:
Okay, now I'm truly lost. This is the error on line 22 -
```
##[error] [-] Discovery in D:\a\_work\1\s\tests\GitDsc\GitDsc.Tests.ps1 failed with:
##[error] System.Management.Automation.ActionPreferenceStopException: The running command stopped because the preference variable "ErrorActionPreference" or common parameter is set to Stop: The 'Assert-Git' command was found in the module 'GitDsc', but the module could not be loaded. For more information, run 'Import-Module GitDsc'.
##[error]    at System.Management.Automation.ExceptionHandlingOps.CheckActionPreference(FunctionContext funcContext, Exception exception)
##[error]    at System.Management.Automation.Interpreter.ActionCallInstruction`2.Run(InterpretedFrame frame)
##[error]    at System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame frame)
##[error]    at System.Management.Automation.Interpreter.EnterTryCatchFinallyInstruction.Run(InterpretedFrame frame)
##[error] at <ScriptBlock>, D:\a\_work\1\s\resources\GitDsc\GitDsc.psm1: line 22
##[error] at <ScriptBlock>, C:\Program Files\WindowsPowerShell\Modules\Pester\5.6.1\Pester.psm1: line 3090
##[error] at Invoke-File, C:\Program Files\WindowsPowerShell\Modules\Pester\5.6.1\Pester.psm1: line 3099
##[error] at Invoke-BlockContainer, C:\Program Files\WindowsPowerShell\Modules\Pester\5.6.1\Pester.psm1: line 3008
##[error] at Discover-Test, C:\Program Files\WindowsPowerShell\Modules\Pester\5.6.1\Pester.psm1: line 1480
##[error] at Invoke-Test, C:\Program Files\WindowsPowerShell\Modules\Pester\5.6.1\Pester.psm1: line 2474
##[error] at Invoke-Pester<End>, C:\Program Files\WindowsPowerShell\Modules\Pester\5.6.1\Pester.psm1: line 5110
##[error] at <ScriptBlock>, D:\a\_work\_temp\52fc0625-641c-49ab-b034-dbe3d8f921f3.ps1: line 5
##[error] at <ScriptBlock>, <No file>: line 1
```

But yet the module is imported on line 19 in the `BeforeAll` block ???

Trenly:
Why do I even try.  I uninstalled Git, and the tests passed for me. I tried it in a VM, and the tests passed for me. I have no clue what's going wrong.  I'll revisit this later once I have a bit more time

PR #90, opened 2024-11-02
Add Microsoft.Windows.Assertion Module

Trenly:

### Premise

Until DSC v3 is released, there isn't a good way to assert specific system parameters in DSC - Especially for use with WinGet. This module adds several different DSC Resources that can be used for asserting the state of a system, primarily for use in WinGet configuration files. This will allow DSC users in enterprises to target specific architectures, specific domain roles, computers with specific print servers, etc. by adding these assertions directly into their configuration YAML instead of having to push out a wrapper script that performs the checks before invoking the DSC, and without having to write their own implementations using the `Script` DSCResource.

-----


resources/Microsoft.Windows.Assertion/Microsoft.Windows.Assertion.psm1
```diff
@@ -0,0 +1,461 @@
...
+
+    [void] Set()
+    {
+        # This resource is only for asserting the OS Install Date.
+    }
+}
+
+# This is the same function from Microsoft.Windows.Developer, just included here as it seemed to make sense
+[DSCResource()]
+class OsVersion
```
Gijsreyn, line 218: Should we actually include it? DSC v3 already has osinfo, see: https://github.com/PowerShell/DSC/tree/main/osinfo

resources/Microsoft.Windows.Assertion/Microsoft.Windows.Assertion.psd1
```diff
@@ -0,0 +1,46 @@
...
+        'OsVersion',
+        'CsManufacturer',
+        'CsModel',
+        'CsDomain',
+        'PowerShellVersion',
+        'PnPDevice'
+    )
+    PrivateData          = @{
+        PSData = @{
+            # Tags applied to this module. These help with module discovery in online galleries.
```
Gijsreyn, line 27: niy: Can you both add `LicenseUri` and `ProjectUri`? This helps user discover the project when they're installing it from PowerShell Gallery :)

resources/Microsoft.Windows.Assertion/Microsoft.Windows.Assertion.psm1
Trenly: Not everyone is going to move to DSCv3 right away, and currently WinGet only really supports DSCv2. I know that this resource is / will be duplicated, but I figure that it's easy enough to add a deprecation warning, or perhaps even call the DSCv3 method internally, once DSCv3 is officially released and is more widely supported

resources/Microsoft.Windows.Assertion/Microsoft.Windows.Assertion.psm1
```diff
@@ -0,0 +1,461 @@
...
+
+    [bool] Test()
+    {
+        $currentState = $this.Get()
+        return $currentState.Edition -eq $currentState.RequiredEdition
+    }
+
+    [void] Set()
+    {
+        # This resource is only for asserting the Edition ID requirement.
```
ryfu-msft, line 46: I am wondering if it makes sense to throw an error if this is ever called. That way the user doesn't accidentally think that something was actually set.

resources/Microsoft.Windows.Assertion/Microsoft.Windows.Assertion.psm1
```diff
@@ -0,0 +1,461 @@
...
+    DEGRADED
+    UNKNOWN
+}
+
+[DSCResource()]
+class OsEditionId
+{
+    # Key required. Do not set.
+    [DscProperty(Key)]
+    [string]$SID
```
ryfu-msft, line 20: I would change the mandatory property to become the Key. We don't need an SID anymore. I think this can be applied to all the new DSC resources below.

ryfu-msft (changes requested):

resources/Microsoft.Windows.Assertion/Microsoft.Windows.Assertion.psm1
Trenly: I'm not sure. 

If I'm not mistaken, the way `winget configure` works is that it will test the resource, if it is not in the desired state, it will attempt to set it, then it checks to be sure that it set properly. If setting throws an error, I assume WinGet would be able to catch it?

resources/Microsoft.Windows.Assertion/Microsoft.Windows.Assertion.psm1
Trenly: Will do!

Trenly:
Didn't realize I still have scriptanalyzer errors

resources/Microsoft.Windows.Assertion/Microsoft.Windows.Assertion.psm1
Trenly: It looks like I was mistaken, and that WinGet doesn't actually check that the configuration was applied successfully before moving on. In order for these to work as intended, each resource will have to `throw` when the set method is called

Trenly: I've now added the functionality for it to throw an error when the setter is called along with related tests. Here's some sample output : 

```powershell
PS C:\Users\Trenly> Invoke-DscResource -Name PnPDevice -ModuleName Microsoft.Windows.Assertion -Method Set -Property @{Ensure='Absent';Status='OK'}
OperationStopped: Assertion Failed. One or more PnP devices found which matched the parameters
PS C:\Users\Trenly> Invoke-DscResource -Name PnPDevice -ModuleName Microsoft.Windows.Assertion -Method Set -Property @{Ensure='Present';FriendlyName='Test'}
OperationStopped: Assertion Failed. No PnP devices found which matched the parameters
PS C:\Users\Trenly> Invoke-DscResource -Name OsVersion -ModuleName Microsoft.Windows.Assertion -Method Set -Property @{MinVersion='12.01'}
OperationStopped: Assertion Failed. '10.0.22631' >= '12.01' evaluated to 'False'
PS C:\Users\Trenly> Invoke-DscResource -Name OsInstallDate -ModuleName Microsoft.Windows.Assertion -Method Set -Property @{Before='11/5/2024';After='11/3/2024'}
OperationStopped: Assertion Failed. '11/5/2024' >= '08/05/2024 09:22:31' >= '11/3/2024' evaluated to 'False'
```

And from WinGet -
```
Apply :: OsEditionId [os-edition]
  The configuration unit failed while attempting to apply the desired state.
Assertion Failed. Expected 'Enterprise Two' but received 'Enterprise'
<See the log file for additional details>
```

Trenly:
@ryfu-msft - I'm a bit stuck here and don't know how to go about it. When I run the pester tests locally, I see them pass -
```
PS E:\winget-dsc> Invoke-Pester E:\winget-dsc\tests\Microsoft.Windows.Assertion

Starting discovery in 1 files.
Discovery found 73 tests in 711ms.
Running tests.
[+] E:\winget-dsc\tests\Microsoft.Windows.Assertion\Microsoft.Windows.Assertion.Tests.ps1 5.3s (3.8s|814ms)
Tests completed in 5.33s
Tests Passed: 73, Failed: 0, Skipped: 0, Inconclusive: 0, NotRun: 0
PS E:\winget-dsc> $env:COMPUTERNAME
CORSAIR-JLU545
```

I even went to a completely different computer, checked out the repo, and ran the tests, where they all passed.
```
PS C:\git\winget-dsc> Invoke-Pester C:\git\winget-dsc\tests\Microsoft.Windows.Assertion

Starting discovery in 1 files.
Discovery found 73 tests in 510ms.
Running tests.
[+] C:\git\winget-dsc\tests\Microsoft.Windows.Assertion\Microsoft.Windows.Assertion.Tests.ps1 17.3s (16.44s|347ms)      
Tests completed in 17.3s
Tests Passed: 73, Failed: 0, Skipped: 0, Inconclusive: 0, NotRun: 0
PS C:\git\winget-dsc> $env:COMPUTERNAME
M5423133
```

Then, I used `Invoke-DSCResource` to manually test, and everything passed there; Granted I did have to use valid PnPDevices that were on my system for the cases where I expected it to match, but all of the ones that I expected to not match didn't match.

As a final effort, I opened up a new Windows Sandbox, mapped in the folder from Git, installed PowerShell 7, and tried running the tests. They all passed again -
```
PS C:\Users\WDAGUtilityAccount> Invoke-Pester C:\Users\WDAGUtilityAccount\Desktop\winget-dsc\tests\Microsoft.Windows.Assertion\

Starting discovery in 1 files.
Discovery found 73 tests in 715ms.
Running tests.
[+] C:\Users\WDAGUtilityAccount\Desktop\winget-dsc\tests\Microsoft.Windows.Assertion\Microsoft.Windows.Assertion.Tests.ps1 11.71s (10.33s|699ms)
Tests completed in 11.73s
Tests Passed: 73, Failed: 0, Skipped: 0, Inconclusive: 0, NotRun: 0
PS C:\Users\WDAGUtilityAccount> $env:COMPUTERNAME
C1BE12D1-33D2-4
```

Gijsreyn:
Trenly, maybe this helps:

image

You can run it through Azure Pipelines itself. I had a similar problem, but this was related to a PowerShell module not being available on the hosted agents (see #120). If you want, you can always hit me up to give it a try. Cheers.

Trenly:
> Trenly, maybe this helps:
> 

Are you saying that the module for `Get-PNPDevice` isn't present? Because it seems to me the only reason it's failing is because the mocks are not applying. If the mocks were applying, you wouldn't see the CIM error

Gijsreyn:
> > Trenly, maybe this helps:
> 
> Are you saying that the module for `Get-PNPDevice` isn't present? Because it seems to me the only reason it's failing is because the mocks are not applying. If the mocks were applying, you wouldn't see the CIM error

No, that's not actually what I was referring to. What I wanted to refer to is that I made an assumption that you hadn't tried running the test on an Azure Pipelines hosted agent. I have seen some strange cases now where some things are available on your local machine (or sandbox) but slightly behave differently on the hosted agents. That's why I wanted to run it through the agent itself and see what it was doing. The pipeline running in this repository doesn't always output the verbosity message you sought. And I think the `Get-PnPDevice` was there; otherwise, the other tests would have also failed.

P.S. The tests were run under Pester v5.6.1.

tests/Microsoft.Windows.Assertion/Microsoft.Windows.Assertion.Tests.ps1
```diff
@@ -0,0 +1,552 @@
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License.
+using module Microsoft.Windows.Assertion
+
+$ErrorActionPreference = 'Stop'
+Set-StrictMode -Version Latest
+
+<#
+.Synopsis
+   Pester tests related to the Microsoft.WinGet.Developer PowerShell module.
```
Gijsreyn, line 10: Nit: Should be Microsoft.WinGet.Assertion

Issue #34, opened 2024-05-24
Windows Settings DSC Resources


denelon:
### Description of the new feature / enhancement

I'd like to capture all Windows Settings as DSC Resources.

Microsoft.Windows.Setting would be the "top level" module so all the children could be included for WinGet to be able to `winget configure export`.

This list below is organized based on appearance in Windows Settings (Windows 11).

I'll update the description as we continue to break things out. I just wanted to get this issue in place to help track the work as we coordinate with other collaborators.

Note: This module is in development, and in its current implementation it would generate a large number of entries in the configuration file. We're considering the concept of "export not default" which would essentially only generate lines of configuration for settings that were not in their default state to reduce the size of the WinGet configuration file.

denelon:
Mentioning @stephengillie to see if that let's me assign the issue...

stephengillie:
I can't edit the above comment, so adding new Issues here:

* [ ] Microsoft.Windows.Setting.System
  - #81
* [ ] Microsoft.Windows.Setting.Bluetooth
* [ ] Microsoft.Windows.Setting.Device
* [ ] Microsoft.Windows.Setting.Network
* [ ] Microsoft.Windows.Setting.Personalization
* [ ] Microsoft.Windows.Setting.App
* [ ] Microsoft.Windows.Setting.Account
* [ ] Microsoft.Windows.Setting.Time
* [ ] Microsoft.Windows.Setting.Language
* [ ] Microsoft.Windows.Setting.Gaming
* [ ] Microsoft.Windows.Setting.Accessibility
  - #53
* [ ] Microsoft.Windows.Setting.Privacy
* [ ] Microsoft.Windows.Setting.Security
* [ ] Microsoft.Windows.Setting.Update

Issue #74, opened 2024-10-23
Allow Machine Configuration reasons to be included


Gijsreyn:
### Description of the new feature / enhancement

With WinGet now in preview for Server 2025, using Machine Configuration to set up resources is becoming more attractive.

If DSC resources are build for Machine Configuration, it must have the `Reasons` property configured. For more information, check out the outstanding documentation in the [DSC v2.0](https://learn.microsoft.com/en-us/powershell/dsc/tutorials/create-dsc-resource-machine-config?view=dsc-2.0#add-the-reasons-property) space.

PR #93, opened 2024-11-04
Windows settings update

Gijsreyn:

-----

Addresses the Microsoft.Windows.Setting.WindowsUpdate on #34


Gijsreyn:
@denelon Hello, I thought giving a shot on the Windows Update settings. Althought I came pretty far and I left the PR as draft, there are some oddities I need help with. Hopefully you (or others) can help me out:

1. When setting the delivery optimization settings through the UI, additional registry keys are created:

image

I couldn't find any relevant documentation if these keys are actually required, or if the keys that are being set now, are sufficient.

2. The delivery optimization settings require elevation to be read
3. I attempted to mimic the work of Stephen in the test. Unfortunately, Pester only creates a hive in HKCU. All settings that we attempt to set, are done in HKLM or HKU.

denelon:
@crutkas and I have been working on some of the challenges with Windows update for his [setup](https://github.com/crutkas/setup). There are some new ACL changes associated with some registry keys in newer versions of Windows, and we've been working on the "right" way to do this.

I don't know if you've seen the experimental "configureSelfElevate" feature in WinGet, but if you decorate the configuration file appropriately, you'll get a single UAC when running a configuration with one or more resources requiring elevation. That doesn't really help with the challenges here, but it's worth mentioning.

denelon:
Experimental features in WinGet can be found using `winget features`. If you run `winget features --help` or `winget features -?` you'll get the link to the help document at GitHub.

Yeah, there are still plenty of rough edges leading you to use the Script resource. We're trying to find the examples where that's necessary so we can reason through gaps in implementation or missing DSC Resources.

Secrets are always a challenge when using Configuration as Code. Lots of products default with no password or a well-known password, and it's a bit tricky to ensure they aren't present in configuration files. This is definitely an area we will need to explore further to come up with some best-practices and model them.

I'll reach out to the PowerShell team to see what they are thinking in regard to PSResourceGet and installing modules.

denelon:
Related to:
* #104

Gijsreyn:
Demitrius, can I put this PR on review to get some assistance on writing unit tests?

denelon:
You bet :)

crutkas:
Can’t wait to try this out!

Gijsreyn:
@ryfu-msft Hey champ, can you take a look with your magic eyes and hopefully give me some assistance in writing Pester tests?

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
```diff
@@ -0,0 +1,377 @@
...
+    [nullable[bool]] $AllowAutoWindowsUpdateDownloadOverMeteredNetwork
+
+    [DscProperty()]
+    [nullable[bool]] $RestartNotificationsAllowed
+
+    [DscProperty()]
+    [nullable[bool]] $SmartActiveHoursState
+
+    [DscProperty()]
+    [nullable[int]] $UserChoiceActiveHoursEnd
```
Trenly, line 272: Inatead of having the assert function, just validate the parameter using `[ValidateRange(0,24)]` and maybe `[AllowNull()]`if its needed

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
```diff
@@ -0,0 +1,377 @@
...
+
+    [DscProperty()]
+    [nullable[int]] $UserChoiceActiveHoursStart
```
Trenly, line 275: Inatead of having the assert function, just validate the parameter using `[ValidateRange(0,24)]` and maybe `[AllowNull()]`if its needed

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
```diff
@@ -0,0 +1,377 @@
...
+    [nullable[int]] $DownloadMode
+
+    [DscProperty()]
+    [nullable[int]] $DownloadRateBackgroundBps
+
+    [DscProperty()]
+    [nullable[int]] $DownloadRateForegroundBps
+
+    [DscProperty()]
+    [nullable[int]] $DownloadRateBackgroundPct
```
Trenly, line 288: Inatead of having the assert function, just validate the parameter using `[ValidateRange(0,100)]` and maybe `[AllowNull()]`if its needed

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
```diff
@@ -0,0 +1,377 @@
...
+
+    [DscProperty()]
+    [nullable[int]] $DownloadRateForegroundPct
```
Trenly, line 291: Inatead of having the assert function, just validate the parameter using `[ValidateRange(0,100)]` and maybe `[AllowNull()]`if its needed

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
```diff
@@ -0,0 +1,377 @@
+$global:WindowsUpdateSettingPath = 'HKLM:\SOFTWARE\Microsoft\WindowsUpdate\UX\Settings'
```
Trenly, line 1: These should be initialized like the accessibility module so the actual registry isn't touched during the tests

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
```diff
@@ -0,0 +1,377 @@
...
+    [nullable[int]] $DownloadRateBackgroundPct
+
+    [DscProperty()]
+    [nullable[int]] $DownloadRateForegroundPct
+
+    [DscProperty()]
+    [nullable[int]] $UploadLimitGBMonth
+    
+    [DscProperty()]
+    [nullable[int]] $UpRatePctBandwidth
```
Trenly, line 297: See above

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
```diff
@@ -0,0 +1,377 @@
...
+    param (
+        [Parameter(Mandatory)]
+        [hashtable] $Parameters
+    )
+
+    if ($Parameters.ContainsKey('DownloadRateBackgroundPct') -or $Parameters.ContainsKey('DownloadRateForegroundPct'))
+    {
+        if ($Parameters.ContainsKey('DownloadRateBackgroundBps') -or $Parameters.ContainsKey('DownloadRateForegroundBps'))
+        {
+            Throw "Cannot set both DownloadRateBackgroundPct/DownloadRateForegroundPct and DownloadRateBackgroundBps/DownloadRateForegroundBps"
```
Trenly, line 151: Instead of doing this, you could use the `ParameterSetName` to create two different parameter sets and let PowerShell do the validation

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
```diff
@@ -0,0 +1,377 @@
...
+        }
+    }
+}
+
+function Initialize-WindowsUpdate
```
Trenly, line 156: It feels like this function shouldn’t be exported. I suggest either updating the psd1 to not export it, or somehow moving it to a non-exported namespace. 

In fact, most of the functions don’t seem like they need to be exported, and could even potentially live inside the class as static functions

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psd1
```diff
@@ -0,0 +1,134 @@
...
+    FunctionsToExport    = '*'
+
+    # Cmdlets to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no cmdlets to export.
+    CmdletsToExport      = '*'
+
+    # Variables to export from this module
+    VariablesToExport    = '*'
+
+    # Aliases to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no aliases to export.
+    AliasesToExport      = '*'
```
Trenly, line 81: Are any aliases actually being exported?

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
Gijsreyn: Both HKLM and HKU are not supported in Pester :(

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
Gijsreyn: I agree `ParameterSetName` would be great for functions. However, it doesn't work for classes (as I'm aware of). The functions are called through the class and there isn't really away to determine it in the class what the user has defined. That's why I'm validating it "later", rather than earlier ...

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
Gijsreyn: Mhmmm. I might have to adjust this in the `New-DscResourceModule`. I have overlooked this one. Nothing has to be exported, expect the `DscResourcesToExport`. I'll update them all.

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
Trenly: According to the [documentation](https://learn.microsoft.com/en-us/powershell/dsc/concepts/class-based-resources?view=dsc-2.0#validation-property-attributes) any validation atteibute should work

Gijsreyn: I might overlook things, but how is this going to work:

```powershell
[DscProperty()]
[Parameter(ParameterSetName = 'ByRate')]
[nullable[int]] $DownloadRateForegroundBps
```

Trenly: Now that is interesting . . . 

I checked out this PR locally and attempted to use the parameter set names myself. Even though the documentation *says* they should work, they absolutely do not. Normally it would throw an error that the sets are mutually exclusive, but it didn't, even when I double checked that my ErrorActionPreference was set to stop. 

I then tried to see if this could be done using a `ValidateScript` - but that gives the error that the attribute argument must be a constant. So either I'm doing something wrong, or not all of the validation attributes are actually supported on Class-based DSC Resources

Gijsreyn: It was indeed. You nailed it by at least trying. I didn't put in the effort because, well... I guess in my head it didn't seem logical to me, nor have I seen it in other implementation.

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
Gijsreyn: Mhm... I didn't think about that. The only thing why I did such implementation, was of issue #74. Despite not implementing it, if `[Reasons]` is the way forward for these DSC resources, we should provide _reasons_ back. 

This allows for Machine Configuration to capture the actual _reason_ why resources couldn't be applied in the desired state we want, instead of throwing errors. The downside of implementing `[Reasons]` in the DSC resources, is more lines of code. Whether that's a good or bad thing, I'll leave that in the air. 

On that note, I guess the road forward is still far away. I wanted to throw it in there if you didn't know about it :)

Thanks for all the reviews lately Kaleb.

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
Trenly: Are you sure? It looks like it was working for the UAC control in #42 - https://github.com/microsoft/winget-dsc/blob/61d2ebf668659f0d3725e171d1c5db8ecbbe93be/resources/Microsoft.Windows.Developer/Microsoft.Windows.Developer.psm1#L152-L160

resources/Microsoft.Windows.Setting.WindowsUpdate/Microsoft.Windows.Setting.WindowsUpdate.psm1
Trenly: Theres no "reason" that you couldn’t do both?  If I'm reading the walkthrough you linked correctly, they have the `ValidateRange` on the `UpdateFrequency` parameter and then later implement `[Reasons]`. I think they are both compatible since the `ValidateRange` only validates what the user input as a parameter, while `[Reasons]` explains why the get method has the results it has

Gijsreyn: You are right, I didn't see that range. I think they are indeed compatible, it is just how we catch it. The difficult part is in the new authoring experience for V3. IntelliSense features are not available (yet). I'll adjust it and remove the functions.

.github/actions/spelling/allow.txt
```diff
@@ -37,3 +37,10 @@ uilt
 Windo
 ELSPROBLEMS
 requ
+whatif
+pscustomobject
+itsdangerouss
+itsdangerouss
+Bandwith
+PCs
+wmiprvse
```
Trenly: This one should probably be in `expect/software.txt`

You don't have to change it, it's just a preference to try and keep things organized by what type of word they are. I'll have to update the advice file to give better guidance, but the way it's structured is that `expect` is for "real" words / terms that appear semi-frequently, and `allow` is for the weird one-offs like `itsdangerouss`

denelon:
It looks like there are conflicts in this PR now. If you're still working on it, it might be best to convert it to a draft, but if you're ready for review, keep it open after you fix the conflicts :)

Gijsreyn:
Oops. I made a mistake. It should be fixed now. There are also tests added now and while adding the test, I found two caveats, which were quite interesting. If there are any suggestions on it to do it differently, then please let me know.

Originally I designed the validation checking within a function. This was replaced after the suggestion of Kaleb: [7ed2c35](https://github.com/microsoft/winget-dsc/pull/93/commits/7ed2c35be6b4dd7a3dceebdde8f511e35d16446f). Here came the challenge: when performing testing, when the validation is not within the range of of zero, it cannot grab the original state. That's why I added an additional check in when that particalur property is set. Second challenge was the fact that when specifying either Pct/Bps, it throws an error. Now I'm simply removing the parameters. I'm going to be honest here, there is some rework in here which I totally miscalculate. If either set is provided as properties, it should remove the others, otherwise both settings don't work.

Because the pull request is already quite large, I hope it can be pulled in. I'll fix this one later and add it as TODO on #104 . If not, please let me know.

Gijsreyn:
Hey @ryfu-msft . I was wondering if you could take a look at this one after my last remark and either approve or let me make some changes. Please let me know.

Issue #100, opened 2024-11-04
Microsoft.Windows.Setting.Time


denelon:
### Description of the new feature / enhancement

Windows Settings for Time

Gijsreyn:
@denelon Working on this one! Thanks for creating the issue.

Gijsreyn:
TODO: Additional clocks.

PR #109, opened 2024-11-05
Windows settings time

Gijsreyn:

-----

Addresses issue #34 and #100


Gijsreyn:
@ryfu-msft And this one. There is still an issue that I have opened at [PSDesiredStateConfiguration](https://github.com/PowerShell/PSDesiredStateConfiguration/issues/125).

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,235 @@
+$global:tzAutoUpdatePath = "HKLM:\SYSTEM\CurrentControlSet\Services\W32Time\Parameters"
```
Trenly, line 1: Is there a reason for not initializing these like the registry paths are initialized in `Microsoft.Windows.Accessibility` so that tests don’t use the real registry?

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,235 @@
...
+
+    This example tests whether the time zone is set to Pacific Standard Time.
+#>
+[DscResource()]
+class Time 
+{
+    # Timezone values are taken from the list of timezones (Get-TimeZone -ListAvailable).Id
+    # TODO: Track issue 125 on PSDesiredStateConfiguration repository to add a ValidateSet for time zones
+    [DscProperty(Key)]
+    # [ValidateSet(
```
Trenly, line 78: Should probably be a validate script; Ref your linked issue - https://github.com/PowerShell/PSDesiredStateConfiguration/issues/125#issuecomment-2456396017

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,235 @@
...
+    #     "AUS Central Standard Time", "E. Australia Standard Time", "AUS Eastern Standard Time", "West Pacific Standard Time", 
+    #     "Tasmania Standard Time", "Vladivostok Standard Time", "Lord Howe Standard Time", "Bougainville Standard Time", 
+    #     "Russia Time Zone 10", "Magadan Standard Time", "Norfolk Standard Time", "Sakhalin Standard Time", "Central Pacific Standard Time", 
+    #     "Russia Time Zone 11", "New Zealand Standard Time", "UTC+12", "Fiji Standard Time", "Kamchatka Standard Time", 
+    #     "Chatham Islands Standard Time", "UTC+13", "Tonga Standard Time", "Samoa Standard Time", "Line Islands Standard Time"
+    # )]
+    [string] $TimeZone = (Get-TimeZone).Id
+
+    [DscProperty()]
+    [TimeZoneAutomatically] $SetTimeZoneAutomatically = [TimeZoneAutomatically]::NTP
```
Trenly, line 113: Why an enum? Couldn’t this just be a boolean?

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,235 @@
...
+        $currentState.SetTimeZoneAutomatically = [Time]::GetTimeZoneAutoUpdateStatus()
+        $currentState.TimeZone = (Get-TimeZone).Id
+        $currentState.ShowSystemTrayDateTime = [Time]::GetShowSystemTrayDateTimeStatus()
+
+        return $currentState
+    }
+
+    [void] Set()
+    {
+        if ($this.Test())
```
Trenly, line 133: I appreciate attempting to return early, but doesn’t this result in two calls to `$this.get()`?

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,235 @@
...
+        }
+
+        if ($currentState.TimeZone -ne $this.TimeZone)
+        {
+            Set-TimeZone -Id $this.TimeZone    
+        }
+
+        if ($currentState.ShowSystemTrayDateTime -ne $this.ShowSystemTrayDateTime)
+        {
+            $desiredState = switch ($this.ShowSystemTrayDateTime)
```
Trenly, line 152: Could you just cast to an int instead of using the switch statement?
```
$desiredState = [int]$this.ShowSystemTrayDateTime)
```

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,235 @@
...
+        else
+        {
+            return ($keyValue -as [TimeZoneAutomatically])
+        }        
+    }
+
+    static [bool] GetShowSystemTrayDateTimeStatus()
+    {
+        $value = TryGetRegistryValue -Key $global:SysTrayPath -Property ([Time]::ShowSystemTrayDateTimeProperty)
+        if (([string]::IsNullOrEmpty($value)) -or ($null -eq $value))
```
Trenly, line 208: You shouldn’t need to check for `($null -eq $value)` since it is already checked in the string method

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: Please see the comment on #93. Pester docs: https://pester.dev/docs/usage/testregistry

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: I don't know what it was, but somehow it had odd behaviours when doing so. is it nit or does it needs to be adjusted?

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: Clever thinking, will adjust!

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: You are right! Modified.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: It does indeed. I tried figuring out a way to get the current state early on without having multiple calls. I don't want to compare apples with apples, but the same behaviour was happening in `Microsoft.Windows.Setting.Accessibility`:

```powershell
if ($this.Test())
{
    return
}

if (-not (Test-Path -Path $global:MagnifierRegistryPath))
{
    New-Item -Path $global:MagnifierRegistryPath -Force | Out-Null
}

if ($this.Magnification -ne [MagnificationValue]::KeepCurrentValue)
{
    $desiredMagnification = switch ([MagnificationValue]($this.Magnification))
    {
        None { 0 }
        Low { 100 }
        Medium { 200 }
        High { 300 }
    }

    Set-ItemProperty -Path $global:MagnifierRegistryPath -Name $this.MagnificationProperty -Value $desiredMagnification -Type DWORD
}

$currentState = $this.Get()
```

If you have a better way, please shoot Kaleb :)

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: The only thing that I can think of, is create a enum without the spaces, and add another helper function to translate it. Might be a bit overkill, so I'm curious what the PS team is going to mention.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Trenly: No worries, was just curious

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Trenly: Will test on my machine after work - I have a few ideas I need to check

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Trenly: Interesting. . .

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Trenly: So a couple things - On both of my machines, updating this registry key changes the value of `Set time automatically` and not `Set time zone automatically`

Then, when I checked out the PR without any modifications attempting to use the resource failed to set the desired state -
```
PS C:\Users\Trenly> Invoke-DscResource -Name Time -ModuleName Microsoft.Windows.Setting.Time -Method Set -Property @{SetTimeZoneAutomatically = 'NoSync'}

RebootRequired
--------------
         False

PS C:\Users\Trenly> Invoke-DscResource -Name Time -ModuleName Microsoft.Windows.Setting.Time -Method Get -Property @{}

TimeZone              SetTimeZoneAutomatically ShowSystemTrayDateTime
--------              ------------------------ ----------------------
Central Standard Time                      NTP                  False
```

After looking at the registry more, I see why you chose to use the enum, but the registry key is really only acting as a boolean. I did some investigation and found that if the key is not present, the key is empty, or the key has *any value* other than `NoSync` (case insensitive), then NTP is enabled and the time will be set automatically.  

My recommendation would be to have the parameter be a nullable bool, and then add two hidden strings so you can do something like this:
```
static hidden [string] $NtpEnabled= 'NTP'
static hidden [string] $NtpDisabled= 'NoSync'
. . .
if ($currentState.SetTimeAutomatically -ne $this.SetTimeAutomatically)
{
	$desiredState = $this.SetTimeAutomatically ? [Time]::NtpEnabled : [Time]::NtpDisabled
	Set-ItemProperty -Path $global:tzAutoUpdatePath -Name ([Time]::SetTimeZoneAutomaticallyProperty) -Value $desiredState
}
```

That way the user is presented with a boolean option both in settings and in DSC. However, we can wait for @ryfu-msft or @denelon to give their opinions

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,231 @@
...
+        {
+            Set-ItemProperty -Path $global:tzAutoUpdatePath -Name ([Time]::SetTimeZoneAutomaticallyProperty) -Value $this.SetTimeZoneAutomatically
+        }
+
+        if ($currentState.TimeZone -ne $this.TimeZone)
+        {
+            Set-TimeZone -Id $this.TimeZone    
+        }
+
+        if ($currentState.ShowSystemTrayDateTime -ne $this.ShowSystemTrayDateTime)
```
Trenly, line 150: You should be checking if the user actually requested to set these. I invoked the DSC resource without setting this property and the clock in my tray disappeared

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psd1
```diff
@@ -0,0 +1,131 @@
...
+    CmdletsToExport      = '*'
+
+    # Variables to export from this module
+    VariablesToExport    = '*'
+
+    # Aliases to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no aliases to export.
+    AliasesToExport      = '*'
+
+    # DSC resources to export from this module
+    DscResourcesToExport = @('Time')
```
Trenly, line 84: If you're not going to add it in this PR, once this PR merges I would request that you add a follow up issue to track the addition of adding clocks for different timezones under `HKEY_CURRENT_USER\Control Panel\TimeDate\AdditionalClocks\1`

image

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psd1
```diff
@@ -0,0 +1,131 @@
...
+    FunctionsToExport    = '*'
+
+    # Cmdlets to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no cmdlets to export.
+    CmdletsToExport      = '*'
+
+    # Variables to export from this module
+    VariablesToExport    = '*'
+
+    # Aliases to export from this module, for best performance, do not use wildcards and do not delete the entry, use an empty array if there are no aliases to export.
+    AliasesToExport      = '*'
```
Trenly, line 81: Looks like exports need to be fixed on this PR too

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,231 @@
...
+            {
+                $parameters[$property.Name] = $property.Value
+            }
+        }
+
+        return $parameters
+    }
+    #endRegion Time helper functions
+}
+#endRegion Classes
```
Trenly, line 231: nit: Newline at end of file

Trenly (changes requested):

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psd1
Gijsreyn: I'll put it on #109

Gijsreyn: @Trenly I have partially started on this one. Do you mind sharing your thoughts on #100 how we are going to set both 1 and 2 in the registry?

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: I have added this, can you verify once more?

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psd1
Trenly: I would implement it as a single `AdditionalClock` resource, which has Properties of `Ensure`, `TimeZone`, `DisplayName`. 

For `Get` -
* Fetch the values in `1`
* Compare against properties
* If the properties match, set ensure present and return the current state
* If the properties don't match, fetch the values in `2`
* Compare against properties
* If the properties match, set ensure present; otherwise, set ensure absent
* Return the current state

For `Test` - 
* Run `Get`
* Compare `Ensure` of the current state to `Ensure` of the desired state
* Return result of comparison

For `Set` it's a bit more complicated, I'd imagine you'd first want to determine if either `1` or `2` had a matching `DisplayName`, if it does, set the values for that clock. If no display names match, check if either timezone matches, if it does, set the values for that clock. If neither are a match, put it in slot 1 if the enable property of the slot is not set. If slot 1 is already enabled, put it in slot 2 if the enable property of the slot is not set. If slot 2 is already enabled, throw an error that the maximum number of additional clocks have been set.

-----

Alternatively, you could just add a parameter for `Id` and restrict it to being `1` or `2` and have it be mandatory for `Set` operations but optional for `Get` and `Test`.  That's honestly probably the better option

Gijsreyn: Thanks for the details. One small question, how will users be able to specify both 1 and 2 in a single configuration document?

Trenly: They call the resource twice - once for setting slot one, and a second time for setting slot two

Gijsreyn: I lack the knowledge on this part. Either clock 1 or clock 2 can be turned on, or both. Shouldn't the user be able to say something like: `AdditionalClock1Settings` and `AdditionaClock2Settings`?

Trenly: No need; They can confugure like:
```Yaml
    - resource: Microsoft.Windows.Setting.Time/AdditionalClock
      directives:
        description: Set clock 1
        allowPrerelease: true
      settings:
        Slot: 1
        DisplayName: UTC
        Timezone: UTC
        Ensure: Present
    - resource: Microsoft.Windows.Setting.Time/AdditionalClock
      directives:
        description: Set clock 2
        allowPrerelease: true
      settings:
        Slot: 2
        DisplayName: Current time in Redmond
        Timezone: Pacific Standard Time
        Ensure: Present
```

Gijsreyn: Ah cool. I'll pick it up after this is merged!

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,376 @@
...
+    [TimeZoneTable] $TimeZone = ((Get-TimeZone).Id -replace '[\+\s\-\(\)\.]', '')
+
+    [DscProperty()]
+    [nullable[bool]] $SetTimeZoneAutomatically
+
+    [DscProperty()]
+    [nullable[bool]] $ShowSystemTrayDateTime
+
+    [DscProperty()]
+    [nullable[bool]] $NotifyClockChange
```
ryfu-msft, line 252: I feel that these should all be separate DSC resources. I think a user shouldn't have to specify the timezone if they only want to Show the datetime in the system tray.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,376 @@
...
+            Write-Verbose "Property `"$($Property)`" could not be found."
+        }
+    } else {
+        Write-Verbose 'Registry key does not exist.'
+    }
+}
+#endRegion Functions
+
+#region Enums
+enum TimeZoneTable {
```
ryfu-msft, line 32: This enum seems like overkill. I would just let the time zone id be used directly as a string parameter instead of creating this enum.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,376 @@
...
+    if ($trimmedVersion -contains $TimeZone) {
+        if ($NoValid.IsPresent) {
+            return $TimeZone
+        }
+
+        return $list[$trimmedVersion.IndexOf($TimeZone)]
+    } else {
+        throw 'Invalid time zone. Please provide a valid time zone without spaces and special characters.'
+    }
+}
```
ryfu-msft, line 202: This can be simplified by just checking if the timezoneId provided is valid by calling `Get-TimeZone -Id <timezoneid>` and seeing if that returns a valid response.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,376 @@
...
+.EXAMPLE
+    PS C:\> Invoke-DscResource -Name Time -Method Test -Property @{ TimeZone = "Pacific Standard Time"}
+
+    This example tests whether the time zone is set to Pacific Standard Time.
+#>
+[DscResource()]
+class Time {
+    # TODO: Track issue 125 on PSDesiredStateConfiguration repository to add a ValidateSet for time zones
+    [DscProperty(Key)]
+    [TimeZoneTable] $TimeZone = ((Get-TimeZone).Id -replace '[\+\s\-\(\)\.]', '')
```
ryfu-msft, line 243:
```suggestion
    [string] $TimeZoneId
```

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,376 @@
...
+
+    This example gets the current time settings on the machine.
+
+.EXAMPLE
+    PS C:\> Invoke-DscResource -Name Time -Method Test -Property @{ TimeZone = "Pacific Standard Time"}
+
+    This example tests whether the time zone is set to Pacific Standard Time.
+#>
+[DscResource()]
+class Time {
```
ryfu-msft, line 240:
```suggestion
class TimeZone {
```

ryfu-msft (changes requested):

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: Separated into own class `Clock`.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: Done! The effect without it, is that users don't get the IntelliSense. Docs should guide them.

Gijsreyn:
@ryfu-msft And the same for this one :)

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,307 @@
...
+
+.EXAMPLE
+    PS C:\> Invoke-DscResource -Name Time -Method Get -Property {}
+
+    This example gets the current time settings on the machine.
+#>
+[DscResource()]
+class TimeZone {
+    [DscProperty(Key)]
+    [string] $TimeZone
```
ryfu-msft, line 74: nit: I feel that this should be `TimeZoneId` or `Id` because that is what I see from the output of `Get-TimeZone`

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,307 @@
...
+
+    [DscProperty()]
+    [nullable[bool]] $SetTimeZoneAutomatically
```
ryfu-msft, line 77: I think a separate DSC for this field specifically. It doesn't makes sense to have to specify a timeZone (key) in order to be able to set this value. Also this value has no effect until you give it a manual time to set it to.

tests/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.Tests.ps1
```diff
@@ -0,0 +1,101 @@
...
+
+        Invoke-DscResource -Name TimeZone -ModuleName Microsoft.Windows.Setting.Time -Method Set -Property $desiredState
+
+        $finalState = Invoke-DscResource -Name TimeZone -ModuleName Microsoft.Windows.Setting.Time -Method Test -Property @{}
+        $finalState.InDesiredState | Should -Be $true
+    }
+
+    It 'Set automatic time zone not synchronize' {
+        $object = [TimeZone]::new()
+        $object.SetTimeZoneAutomatically = 'NoSync'
```
ryfu-msft, line 46: Should this take in a boolean value? Regardless, this test can probably be removed if you go with my suggestion from above.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,307 @@
...
+    # helper function for Pester tests
+    [hashtable] ToHashTable() {
+        $parameters = @{}
+        foreach ($property in $this.PSObject.Properties) {
+            if (-not ([string]::IsNullOrEmpty($property.Value))) {
+                $parameters[$property.Name] = $property.Value
+            }
+        }
+
+        return $parameters
```
ryfu-msft, line 169: I would remove this since it's not being used by the tests. I'd rather be explicit with the args we use in tests anyways since that's how they will be invoked.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,307 @@
...
+    }
```
ryfu-msft, line 303: Remove this as well

resources/Help/Microsoft.Windows.Setting.Time/TimeZone.md
```diff
@@ -0,0 +1,45 @@
...
+
+# Time
+
+## SYNOPSIS
+
+This `Time` DSC Resource allows you to manage the time zone, automatic time zone update, and system tray date/time visibility settings on a Windows machine.
+
+## DESCRIPTION
+
+This `Time` DSC Resource allows you to manage the time zone, automatic time zone update, and system tray date/time visibility settings on a Windows machine.
```
ryfu-msft, line 18:
```suggestion
title: TimeZone
---

# TimeZone

## SYNOPSIS

This `TimeZone` DSC Resource allows you to manage the time zone, automatic time zone update, and system tray date/time visibility settings on a Windows machine.

## DESCRIPTION

This `TimeZone` DSC Resource allows you to manage the time zone, automatic time zone update, and system tray date/time visibility settings on a Windows machine.
```

Probably need to update the descriptions as well since this only relates to `TimeZone`

resources/Help/Microsoft.Windows.Setting.Time/TimeZone.md
```diff
@@ -0,0 +1,45 @@
...
+| `SetTimeZoneAutomatically` | Optional      | Boolean      | Whether to set the time zone automatically. The value should be a boolean. You can find the setting in `Settings -> Time & Language -> Date & Time -> Set time automatically.                                       |     `$true`, `$false`                                                              |
+| `AdjustForDayLightSaving`  | Optional      | Boolean      | Whether to adjust for daylight saving time. The value should be a boolean. You can find the setting in `Settings -> Time & Language -> Date & Time -> Adjust for daylight saving time automatically. | `$true`, `$false` |
+
+
+## EXAMPLES
+
+### EXAMPLE 1 - Set time zone to Pacific Standard Time
+
+```powershell
+Invoke-DscResource -Name Time -Method Set -Property @{ TimeZone = "Pacific Standard Time"}
```
ryfu-msft, line 34: You can also consider including `Get-TimeZone -ListAvailable` as a way to retrieve all available time zones

ryfu-msft (changes requested):

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: Changed.

tests/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.Tests.ps1
Gijsreyn: I removed this test.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: The reason for adding this, is that it is _actually_ used in the test. It can be a bit annoying for users when they run them, that the system gets modified. I'm using this as helper function to get the _current_ state before running the test and after they've run, put back the original state. If you've other recommendations, please let me know.

resources/Help/Microsoft.Windows.Setting.Time/TimeZone.md
Gijsreyn: I didn't fully get what you mean with this?

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -0,0 +1,307 @@
...
+    )
+
+    try {
+        $timeZoneId = (Get-TimeZone -Id $TimeZone -ErrorAction Stop).Id
+    } catch {
+        throw [System.Configuration.ConfigurationException]::new("Executing 'Get-TimeZone' failed. Error: $($PSItem.Exception.Message)")
+    }
+
+    return $timeZoneId
+}
```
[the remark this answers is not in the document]
Gijsreyn: After rechecking your comment multiple times and looking at the code, I can see why I did it. Before this function, I was _enumming_ everything to allow users to always fill in a proper ID. I moved it to this, because it doesn't really make sense to test either `$true` or `$false` on this. If a user inserts an invalid ID, it should basically throw _somewhere_. Perhaps we can leave this function alone and let it throw when insert an invalid ID. In that way, when `Invoke-DscResource` is called, it returns a proper message. Let me know what  you think.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: After debugging for countless of hours, I got the confusing. Let me try to eleborate what is adjusted now. Imagine the following:

image

Both settings related to time, so I can recon that it belongs to the `TimeZone` class. I have now updated both properties. The situation is now as followed:

1. When a user sets the **SetTimeZoneAutomatically** to `$true`, it overrules all others even when the user sets them
2. When a user leaves it empty and sets the other values, they are set correctly. Here I made a small adjustment for daylight savings, because there are just to many registry keys to be modified. I'm using _tzutil_ to adjust this.

I have tried to follow the principle that @denelon mentioned in #34 .

Please revisit the adjustments once again :)

Gijsreyn:
@denelon I'm adding you to the chain as well because I can remember you making a remark in the past that you were checking with the Windows team on registry keys. When spawning up a fresh machine from DevHome, I noticed something odd that I hadn't seen before.

I just learned about the privacy and security settings, but they bite this PR (a bit). For now, I have added an additional check on the relevant registry keys I could capture when `SetTimeZoneAutomatically` is `$true`, but I'm curious about your thoughts. Should we test this, and if it is disabled, don't we do anything? Or do we just continue?

denelon:
That's a great find!

What settings are getting applied here, and what happens if we don't enable the location permissions?

Gijsreyn:
If the location permissions aren't set, users must manually enter the time.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -39,57 +40,107 @@ function Get-ValidTimeZone {
...
     [DscProperty(Key)]
-    [string] $TimeZone
+    [string] $Id
 
     [DscProperty()]
     [nullable[bool]] $SetTimeZoneAutomatically
 
+    [DscProperty()]
+    [nullable[bool]] $SetTimeAutomatically
+
```
ryfu-msft, line 117: How do you feel about moving things like `SetTimeAutomatically` to a separate DSC resource called `Time`.
When I look at the settings page for TimeZone, I only see `SetTimeZoneAutomatically` and if that is not true, then we have the option to set the TimeZoneId and DaylightSaving. I think this DSC resource should try to replicate what we see in the settings app. I think this is the best way to prevent scope creep for this individual DSC resources.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -39,57 +40,107 @@ function Get-ValidTimeZone {
...
+    static hidden [string] $NtpDisabled = 'NoSync'
+    static hidden [bool] $SupportsDaylightSavingProperty = $false
 
     TimeZone() {
-        $this.TimeZone = (Get-TimeZone).Id
+        $timeZone = Get-TimeZone
+        $this.Id = $timeZone.Id
+
+        # We set the SupportsDaylightSavingProperty to the value that is supported by the current time zone
+        [TimeZone]::SupportsDaylightSavingProperty = $timeZone.SupportsDaylightSavingTime ? $true : $false
```
ryfu-msft, line 133: $timeZone.SupportsDaylightSavingTime is already a boolean so no need to have a ternary operator.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -39,57 +40,107 @@ function Get-ValidTimeZone {
...
     [DscProperty()]
     [nullable[bool]] $AdjustForDaylightSaving
 
-    static hidden [string] $SetTimeZoneAutomaticallyProperty = 'Type'
+    static hidden [string] $SetTimeZoneAutomaticallyProperty = 'Start'
+    static hidden [string] $SetTimeAutomaticallyProperty = 'Type'
     static hidden [string] $AdjustForDaylightSavingProperty = 'DynamicDaylightTimeDisabled'
+    static hidden [string] $NtpEnabled = 'NTP'
+    static hidden [string] $NtpDisabled = 'NoSync'
+    static hidden [bool] $SupportsDaylightSavingProperty = $false
```
ryfu-msft, line 126: This shouldn't be static as this value changes depending on the TimeZone instance. It would be better to have this as a nonconfigurable parameter that we determine in the Get() method. Here's an [example](https://github.com/microsoft/winget-dsc/blob/583d2c7671481e5d52e776e9f5af040e97fe51a0/resources/Microsoft.Windows.Developer/Microsoft.Windows.Developer.psm1#L59-L60)

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -39,57 +40,107 @@ function Get-ValidTimeZone {
...
     [TimeZone] Get() {
         $currentState = [TimeZone]::new()
-        $currentState.SetTimeZoneAutomatically = [TimeZone]::GetTimeZoneAutoUpdateStatus()
-        $currentState.AdjustForDaylightSaving = [TimeZone]::GetDayLightSavingStatus()
+        $currentState.SetTimeZoneAutomatically = [TimeZone]::GetTimeZoneAutomaticallyStatus()
+        $currentState.SetTimeAutomatically = [TimeZone]::GetTimeAutomaticallyStatus()
+
+        if ($currentState::SupportsDaylightSavingProperty) {
+            $currentState.AdjustForDaylightSaving = [TimeZone]::GetDayLightSavingStatus()
+        }
```
ryfu-msft, line 143: This value is not dependent on whether the time zone supports daylight saving or not. This should just be a registry check.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -39,57 +40,107 @@ function Get-ValidTimeZone {
...
+    $registryKeys = @('HKLM:\Software\Microsoft\Windows\CurrentVersion\CapabilityAccessManager\ConsentStore\location', 'HKCU:\Software\Microsoft\Windows\CurrentVersion\CapabilityAccessManager\ConsentStore\location', 'HKCU:\Software\Microsoft\Windows\CurrentVersion\CapabilityAccessManager\ConsentStore\location\windows.immersivecontrolpanel*')
+    foreach ($key in $registryKeys) {
+        $res = Get-ItemProperty -Path $key -Name 'Value' -ErrorAction SilentlyContinue
+        if ($res.Value -ne 'Allow') {
+            # TODO: Or should we throw an error?
+            return $false
+        }
+    }
+
+    return $true
```
ryfu-msft, line 55: I feel that this should be its own separate DSC resource for just Location. I'm sure there are other settings that will depend on this, not just Time/TimeZone.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
```diff
@@ -101,55 +152,82 @@ class TimeZone {
...
+            $desiredState = @{
+                Id     = $this.Id
+                Enable = $this.AdjustForDaylightSaving
+            }
 
-            Set-ItemProperty -Path $global:timeZoneInformationPath -Name ([TimeZone]::AdjustForDaylightSavingProperty) -Value $desiredState
+            if ($this.AdjustForDaylightSaving -ne $currentState.AdjustForDaylightSaving) {
+                # Falling back on tzutil because the amount of registry keys to modify is too high, plus difficult to determine which one to modify
+                Set-DaylightSavingTime @desiredState
+            }
```
ryfu-msft, line 182: What happens if you try to enable daylight saving using tzutil on an id that doesn't support daylight saving

ryfu-msft (changes requested):

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: The function `Set-DaylightSavingTime` checks the LASTEXITCODE. IF it not equals 0, it fails and throws an error.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: I didn't totally get you here. It is checking the registry with the `GetDayLightSavingStatus()` method?

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: That's precisely why I did it into _one_ DSC resource. They bite each other. Even if we separate them both, we have to check what options are already enabled in both cases. If we set the Timezone to automatically, we _might_ need to clean up the manual settings and vice versa.

resources/Microsoft.Windows.Setting.Time/Microsoft.Windows.Setting.Time.psm1
Gijsreyn: Partially, as it affects the TimeZone setting, what's your suggestion? A custom module to inherit those functions?

stephengillie: > Partially, as it affects the TimeZone setting, what's your suggestion? A custom module to inherit those functions?

Pardon my jumping into the conversation, but this felt like a good place. It seems to me that a number of these functions should be in a common PowerShell module (or similar location) of some kind. Maybe not so much this location permission testing function, as much as the Registry functions above. It seems like an unnecessary duplication to have every DSC resource independently define how to retrieve values from the Registry. 

What do you think about some kind of centralization for common code?

Gijsreyn: @stephengillie Welcome back and of course you can jump into the conversation. I fully agree with it. I think the DSC community does it also with the [DscResource.Common](https://github.com/dsccommunity/DscResource.Common) module.

If you have any thoughts about implementing it, or want to have a discussion around it, I'm always open to it.

PR #122, opened 2024-11-07
Markdownlint GitHub

Gijsreyn:

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-dsc/pull/122)

### DIFF
--- a/.github/workflows/markdownLint.yaml
+++ b/.github/workflows/markdownLint.yaml
@@ -18,6 +18,7 @@ on:
 
 jobs:
   lint:
+    name: Check Markdown linting
     runs-on: ubuntu-latest
     steps:
     - uses: actions/checkout@v4
--- a/.github/workflows/markdownLint.yaml
+++ b/.github/workflows/markdownLint.yaml
@@ -1,4 +1,20 @@
-on: [push, pull_request]
+name: Markdown Lint
+
+on:
+  push:
+    branches:
+      - '**'
+    tags-ignore:
+      - '**'
+  pull_request_target:
+    branches:
+      - '**'
+    tags-ignore:
+      - '**'
+    types:
+      - 'opened'
+      - 'reopened'
+      - 'synchronize'
 
 jobs:
   lint:
--- a/.github/workflows/markdownLint.yaml
+++ b/.github/workflows/markdownLint.yaml
@@ -4,6 +4,7 @@ on:
   push:
     branches:
       - '**'
+      - '!main'
     tags-ignore:
       - '**'
   pull_request_target:
--- a/.github/workflows/markdownLint.yaml
+++ b/.github/workflows/markdownLint.yaml
@@ -1,0 +1,17 @@
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v4
+      with:
+        fetch-depth: 0
+    - uses: tj-actions/changed-files@v45
+      id: changed-files
+      with:
+        files: '**/*.md'
+    - uses: DavidAnson/markdownlint-cli2-action@v17
+      if: steps.changed-files.outputs.any_changed == 'true'
+      with:
+        globs: ${{ steps.changed-files.outputs.all_changed_files }}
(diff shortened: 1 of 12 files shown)


Gijsreyn:
@denelon

.github/workflows/markdownLint.yaml
```diff
@@ -0,0 +1,34 @@
+name: Markdown Lint
+
+on:
+  push:
```
Trenly, line 4: Will this run the linter twice? Once when a PR is opened and once when it's merged into main?  If so, we should probably ignore pushes to `main`

.github/workflows/markdownLint.yaml
```diff
@@ -0,0 +1,34 @@
...
+    branches:
+      - '**'
+    tags-ignore:
+      - '**'
+  pull_request_target:
+    branches:
+      - '**'
```
Trenly, line 11: Should we be running on every pull request, or should it be only those that target into `main` ? I wouldn't expect there would be many that aren't going into main

Trenly (commented):
Can you help me understand how this action works? Is it like the spell checking where it fails if the linting doesn't match, or is it something that actually goes in and fixes the linting?

Gijsreyn:
> Can you help me understand how this action works? Is it like the spell checking where it fails if the linting doesn't match, or is it something that actually goes in and fixes the linting?

The same as the spell checker. I didn't explicitly set the continueOnError, because $false is the default behaviour.

.github/workflows/markdownLint.yaml
Gijsreyn: I guess only for the PRs, the same as spell checker?

denelon:
I saw some of the "security" related items in one of the files. We want folks to submit security related concerns through the official Microsoft link mentioned in SECURITY.md.

Gijsreyn:
> I saw some of the "security" related items in one of the files. We want folks to submit security related concerns through the official Microsoft link mentioned in SECURITY.md.

I didn't fully get what you mean Demitrius. You mind elaborating a bit more what you sawa and what might needs to be changed?

denelon:
> I didn't fully get what you mean Demitrius. You mind elaborating a bit more what you sawa and what might needs to be changed?

I reviewed it, I think it's fine.

Gijsreyn:
@denelon Thanks! @ryfu-msft If you're also happy, you mind pulling it in?

ryfu-msft:
> Ryan, I don't know why that prerelease is failing, whereas other runs it is successful. No file has been touched on tests. Can we create a separate issue for it for investigation and pull this in?

Lets try one more time and see if its a one off.

Gijsreyn:
At least from my side:
image

Gijsreyn:
I see what is going wrong, but I don't know if it is a bug in the DSC resource, or something on the `dotnet.exe` NuGet side. Version 9.0.0 has just been released, and it's picking that one. Can we comment out the test and I'll take a look at it in another issue? We still have one test that test for prerelease, which is PowerShell.

ryfu-msft:
> I see what is going wrong, but I don't know if it is a bug in the DSC resource, or something on the `dotnet.exe` NuGet side. Version 9.0.0 has just been released, and it's picking that one. Can we comment out the test and I'll take a look at it in another issue? We still have one test that test for prerelease, which is PowerShell.
> 
> EDIT: When I run `dotnet tool install dotnet-ef --prerelease --global`, I would assume it is going to install: [9.0.0-rc.2.24474.1](https://www.nuget.org/packages/dotnet-ef/9.0.0-rc.2.24474.1), but it doesn't.
> 
> EDIT 2: Looking at the description, not really:
> 
> image
> 
> @ryfu-msft Would the `PreRelease` property then make sense, or just instruct users if they want to install a rc for example, just add the full version?

Seems like --prerelease only works if there is a prerelease version that exists later the latest stable release. I would instruct users to resort to using the full version instead of --prerelease since you never know when the next preview release will come out. 

Regarding the tests, it would be better to explicitly point to an older preview version so that the tests don't break in the future when a new release comes out. 

In an ideal world, we should not be using real packages from the nuget gallery but instead stand up our own test package source.

ryfu-msft (approved):

Gijsreyn:
I fully agree with you here Ryan. I'll hit up a small commit in just a second to fix the test and the one for VSCode. If you want, I can remove the --prerelease later. Otherwise, I put it in the docs now. Thanks for checking.

Issue #137, opened 2024-12-09
Applying configuration.dsc.yaml results in error


Gijsreyn:
### Brief description of your issue

On a fresh Hyper-V box created by DevHome, the `configuration.dsc.yaml` inside this repository fails.

### Steps to reproduce

Create a new Hyper-V box through DevHome and execute `winget configure https://raw.githubusercontent.com/microsoft/winget-dsc/refs/heads/main/.configurations/configuration.dsc.yaml` 

### Expected behavior

Apply the configuration

### Actual behavior

The following error occurs:

![Image](https://github.com/user-attachments/assets/f1c7c2d2-8947-4ba2-9d5b-0a685dd87909)

### Environment

```shell
Windows Package Manager v1.9.25200
Copyright (c) Microsoft Corporation. All rights reserved.

Windows: Windows.Desktop v10.0.22621.4460
System Architecture: X64
Package: Microsoft.DesktopAppInstaller v1.24.25200.0

Winget Directories
-----------------------------------------------------------------------------------------------------------------------
Logs                               %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\Diag…
User Settings                      %LOCALAPPDATA%\Packages\Microsoft.DesktopAppInstaller_8wekyb3d8bbwe\LocalState\sett…
Portable Links Directory (User)    %LOCALAPPDATA%\Microsoft\WinGet\Links
Portable Links Directory (Machine) C:\Program Files\WinGet\Links
Portable Package Root (User)       %LOCALAPPDATA%\Microsoft\WinGet\Packages
Portable Package Root              C:\Program Files\WinGet\Packages
Portable Package Root (x86)        C:\Program Files (x86)\WinGet\Packages
Installer Downloads                %USERPROFILE%\Downloads

Links
---------------------------------------------------------------------------
Privacy Statement   https://aka.ms/winget-privacy
License Agreement   https://aka.ms/winget-license
Third Party Notices https://aka.ms/winget-3rdPartyNotice
Homepage            https://aka.ms/winget
Windows Store Terms https://www.microsoft.com/en-us/storedocs/terms-of-sale

Admin Setting                             State
--------------------------------------------------
LocalManifestFiles                        Disabled
BypassCertificatePinningForMicrosoftStore Disabled
InstallerHashOverride                     Disabled
LocalArchiveMalwareScanOverride           Disabled
ProxyCommandLineOptions                   Disabled
[...]

PR #138, opened 2024-12-09
Fix configuration file

Gijsreyn:

-----

Addresses issue #137 
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-dsc/pull/138)

### DIFF
--- a/.configurations/configuration.dsc.yaml
+++ b/.configurations/configuration.dsc.yaml
@@ -60,6 +60,15 @@ properties:
         id: Microsoft.VisualStudioCode
         source: winget
         Ensure: Present
+    - resource: Microsoft.WinGet.DSC/WinGetPackage
+      id: install-powershell
+      directives:
+        description: Install PowerShell
+        allowPrerelease: true
+      settings:
+        id: Microsoft.PowerShell
+        source: winget
+        Ensure: Present
     - resource: Microsoft.VSCode.Dsc/VSCodeExtension
       id: install_vscode-yaml
       dependsOn:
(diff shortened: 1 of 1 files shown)


Gijsreyn:
@Trenly and @ryfu-msft I think there was one small step missing in the configuration file inside this repository. This should fix it.

Trenly:
> @Trenly and @ryfu-msft I think there was one small step missing in the configuration file inside this repository. This should fix it.

The configuration was built presuming the user was already operating inside the cloned repository, as would be the most common workflow when using DevHome

Gijsreyn:
> > @Trenly and @ryfu-msft I think there was one small step missing in the configuration file inside this repository. This should fix it.
> 
> The configuration was built presuming the user was already operating inside the cloned repository, as would be the most common workflow when using DevHome

That was my assumption as well, but for fresh boxes, perhaps it is better to include it?

ryfu-msft (changes requested):
I'd have to agree with Trenly here as well. In my mind the most common workflow would be to start by cloning this repo, then running `winget configure` on the configuration file. Your change will cause it to clone the same repo again. I think it is probably best to leave out the git clone part as that is not really needed within the context of this repo.

Gijsreyn:
Thanks for the sharing and interesting thoughts. I would have assumed the GitClone resource would test for it. If it is already present, it applies the configuration successfully without additional cloning (depending on where the user executes it, of course). 

Regardless, I have removed the GitClone for now. I have fixed up the cloning because it was pointing towards winget-pkgs. I have added the PowerShell requirement.

ryfu-msft (approved):

PR #158, opened 2025-02-16
New module Microsoft.Windows.Settings

Gijsreyn:

-----

This PR introduces the new `Microsoft.Windows.Settings` module. The module will become responsible for configuring a variety of Windows settings without the need for separate individual modules. This addresses the overall modularization, standardization, and introduces the DSC community guidelines.


Gijsreyn:
Hello @denelon and @AmelBawa-msft. As an attempt to address multiple challenges, e.g., script analyzer, modularization, and standardization, this is the main module `Microsoft.Windows.Settings`. I recognize this PR _might_ be quite large, but yes, it's the output Sampler generates. I've explicitly left the `azure-pipelines.yml` file intact, so there's a reference to how it can be built up; as you mentioned earlier, the pipelines were being reviewed. 

I intend to make the module the grandfather of all Windows Settings and slowly move the other modules that have either already been released or are open for pull requests to be added in here. That way, they follow a certain standard, and we don't constantly have to use the same functions repeatedly. Of course, not all registry settings have either 0 or 1. In that case, they will be more self-contained.

denelon:
Thanks! I've got some discussions planned today with @AmelBawa-msft to cover the roadmap for Microsoft.Windows.Setting - I think we may want to stick with the singular name, but I'm open to whichever is best for being idiomatically correct in terms of PowerShell.

Gijsreyn:
Awesome Demitrius. I took the plural name because as with the others that were developed already, it could contain _one_ or _more_ settings. In this case, it will definitely contain more settings. There isn't really a [strongly encouraged development guideline](https://learn.microsoft.com/en-us/powershell/scripting/developer/cmdlet/strongly-encouraged-development-guidelines?view=powershell-7.5) for module names (only for cmdlets, it should be singular), and I also took some reference from the `Az` module (e.g., `Az.Accounts`, `Az.Resources).

I'm always open to rename it to a singular.

denelon:
We're definitely having some challenges thinking about the best way to organize the settings into resources. Given the Windows Settings App get updated and settings get moved around, it's tricky to think about the best way to logically organize them. What ever we decide here will be something we have to live with and support somewhat long term until we get a Windows Settings DSC v3 resource.
I'm also starting some other discussions with internal folks about the pipelines and our publishing process. @AmelBawa-msft has started reviewing the PR and getting up to speed with some of the DSC Community best practices.

resources/Microsoft.Windows.Settings/source/Classes/010.SettingsBase.ps1
```diff
@@ -0,0 +1,114 @@
+<#
```
MattHyman, line 1: Can I ask where the numerical prefixes are coming from on these files? I see other classes like this in DSC community repos and are aware that they may be coming from Sampler templates, but I am not finding a clear explanation or a way to interpret them.

resources/Microsoft.Windows.Settings/source/en-US/Microsoft.Windows.Settings.strings.psd1
```diff
@@ -0,0 +1,17 @@
+<#
+    .SYNOPSIS
+        The localized resource strings in English (en-US) for the
+        resource Microsoft.Windows.Settings module. This file should only contain
+        localized strings for private and public functions.
+#>
+
+ConvertFrom-StringData @'
+    ## Get-RegistryStatus
+    GetRegistryStatusRegistryManualManipulation_ErrorMessage = "The registry path '{0}' is not in the expected state. Entries have been manually modified. Cannot determine the current state."
```
AmelBawa-msft, line 10: I couldn't find usage of some of those names. Do we need to remove them from this file?
- GetRegistryStatusRegistryManualManipulation_ErrorMessage
- GetRegistryStatus_SearchMessage
- GetRegistryStatus_FoundMessage

PR #166, opened 2025-04-01
Updated node.js file and clean up

shakersMSFT:
- [x] I have signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-dsc).
- [ ] This pull request is related to an issue.

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-dsc/pull/166)

PR #169, opened 2025-04-02
Add securityContext: elevated wherever we install git

mdanish-kh:

Git.Git installs in `machine` scope & requires elevation if a user is part of the local admin group. With my suggestion of adding `securityContext: elevated`, I'm making the assumption that most devs on the Desktop installation of Windows will be on the default user account which is part of the local admin group. This means Git would require elevation and install in machine scope

cc @denelon

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-dsc/pull/169)


samples/Repositories/Windows/dev-environment/javascript/nodejs-on-windows/NodeJS.winget
mdanish-kh, line 1: Do we actually need this file as we have the updated one (after PR https://github.com/microsoft/winget-dsc/pull/166) under another directory [samples/Configuration Files/Learn tutorials/JavaScript/NodeJS.winget](https://github.com/microsoft/winget-dsc/blob/main/samples/Configuration%20files/Learn%20tutorials/JavaScript/NodeJS.winget)

mdanish-kh: I figured we don't & removed it in the latest commit

denelon: I think @shakersMSFT wanted to keep those long enough to reason about them as samples.

PR #183, opened 2025-04-25
Add advanced network sharing settings resource to developer provider

MattHyman:
Add advanced network sharing settings resource to developer provider
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-dsc/pull/183)


resources/Microsoft.Windows.Developer/Microsoft.Windows.Developer.psm1
```diff
@@ -800,6 +805,64 @@ class WindowsCapability {
...
     }
 }
+
+[DSCResource()]
+class AdvancedNetworkSharingSetting {
+    [DscProperty(Key, Mandatory)]
+    [AdvancedNetworkSharingSettingName]$Name
+
+    [DscProperty()]
+    [string[]]$Profiles = @()
```
MattHyman, line 815: This is tricky as it intends to disable any enabled profiles if they are not explicitly listed. Can anyone think of a better calling pattern for this? FWIW this is an [unresolved issue on the networking DSC provider](https://github.com/dsccommunity/NetworkingDsc/issues/356).

resources/Microsoft.Windows.Developer/Microsoft.Windows.Developer.psm1
```diff
@@ -800,6 +805,64 @@ class WindowsCapability {
...
+        if (!$this.Test()) {
+            if ($this.Name -eq [AdvancedNetworkSharingSettingName]::NetworkDiscovery) {
+                $group = $this.NetworkDiscoveryGroup
+            } else {
+                $group = $this.FileAndPrinterSharingGroup
+            }
+
+            $firewallGroups = Get-NetFirewallRule -Group $group
+            #Enable
+            $firewallGroups | Where-Object { ($_.Enabled -eq 'False') -and ($this.Profiles -Contains $_.Profile ) } | Set-NetFirewallRule -Enabled True
```
MattHyman, line 859: I ran into a corner case on "File and printer sharing" in which a "private" rule did not exist, causing this logic to not work to enable the setting. I toggled the setting via the UX and then this logic worked as the UX had created the firewall rules.. What is the correct way to enable this scenario via PS without the UX? [ Searching online](https://www.c-sharpcorner.com/article/how-to-enable-or-disable-file-and-printer-sharing-in-windows-102/) makes me wonder if Set-NetFirewallRule on the group directly would do the right thing, or if we need to manually create the firewall rule if it does not initially exist.

MattHyman: I have validated with testing on default firewalls rules that "Set-NetFirewallRule" targeting specific profiles produces the intended result.

resources/Microsoft.Windows.Developer/Microsoft.Windows.Developer.psm1
MattHyman: Based upon the refactor involving Set-NetFirewallRule I believe that this is an acceptable pattern to start with as it will facilitate setting toggling if needed.

MattHyman:
@microsoft/xbox-developer-experience-toolbox for review.

resources/Microsoft.Windows.Developer/Microsoft.Windows.Developer.psm1
```diff
@@ -800,6 +805,64 @@ class WindowsCapability {
...
+        $currentState.Profiles = $this.Profiles
+
+        if ($this.Name -eq [AdvancedNetworkSharingSettingName]::NetworkDiscovery) {
+            $group = $this.NetworkDiscoveryGroup
+        } else {
+            $group = $this.FileAndPrinterSharingGroup
+        }
+
+        $this.EnabledProfiles = Get-NetFirewallRule -Group $group | Where-Object { $_.Enabled -eq 'True' } | Select-Object -Unique -CaseInsensitive -ExpandProperty Profile
+        $currentState.EnabledProfiles = $this.EnabledProfiles
```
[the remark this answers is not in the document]
MattHyman: @guillene, did you try this with the latest version of the code?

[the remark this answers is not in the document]
MattHyman: The intent was to return the list of all profiles that have an enabled rule with the implication that the rule will be "Allow". Are you saying that we need to confirm that for each profile there has at least one enabled rule with no disabled rules? Is there documentation to confirm this?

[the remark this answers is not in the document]
MattHyman: We have reached out to the PS network security team. The network security team has clarified that a firewall rule group is made up of multiple rules, all rules need to be enabled for the group to be seen as enabled. I have updated the PR with this logic.

resources/Microsoft.Windows.Developer/Microsoft.Windows.Developer.psm1
```diff
@@ -800,6 +805,79 @@ class WindowsCapability {
...
+
+        return $currentState
+    }
+
+    [bool] Test() {
+        $currentState = $this.Get()
+
+        # Compare-object is case insensitive by default and does not take null arguments
+        $difference = Compare-Object -ReferenceObject @( $this.Profiles | Select-Object) -DifferenceObject @( $currentState.EnabledProfiles | Select-Object)
```
MattHyman, line 856: There is a bit of a corner case here, if a rule group is in a mixed state and we are trying to enable it then the right things happen. If it is in a mixed state and we are trying to disable it then we ignore it because the machine is in the desire state from a Enabled bool perspective. 

I am happy to write logic to handle this, but for now I believe that this is the right optimization for us so we do not suddenly start changing untargeted profiles for the user. We can approach this as a bug fix if needed.

guillene (approved):
Latest version looks good! 🚀

resources/Microsoft.Windows.Developer/Microsoft.Windows.Developer.psd1
```diff
@@ -35,8 +36,9 @@
                 'PSDscResource_EnableRemoteDesktop',
                 'PSDscResource_EnableLongPathSupport',
                 'PSDscResource_PowerPlanSetting',
-                'PSDscResource_WindowsCapability',
-                'PSDscResource_NetConnectionProfile'
+                'PSDscResource_WindowsCapability'
```
AmelBawa-msft, line 39:
```suggestion
                'PSDscResource_WindowsCapability',
```

AmelBawa-msft (dismissed):

Issue #199, opened 2025-05-09
Field value validation for Microsoft.Windows.Settings


mdanish-kh:
### Brief description of your issue

I can have a configuration file as

```yaml
properties:
  resources:
    - resource: Microsoft.Windows.Settings/WindowsSettings
      directives:
        allowPrerelease: true
        description: Modify Windows Settings
        securityContext: elevated
      settings:
        AppColorMode: SomethingElse # Changes to Light mode
        DeveloperMode: IcanTypeAnythingHere # Enables Developer mode
        SystemColorMode: OrHere # Change to Light mode
        TaskbarAlignment: HereAsWell # Changes to Middle
  configurationVersion: 0.2.0
```

Notice the values of each field under `settings:`

### Steps to reproduce

Run the above configuration file. See the state modified.

### Expected behavior

Even though the input was completely random, the configuration changes the set to "default / fallback behavior" written in the resource.

Default / fallback behavior of AppColorMode is Light mode, DeveloperMode is true, SystemColorMode is Light and TaskbarAlignment is Middle

### Actual behavior

Add input validation in the resource, and don't set anything if the values do not match what we expect

### Environment

```shell
Microsoft.Windows.Settings module version 0.1.0-alpha
```

PR #200, opened 2025-05-09
Change TaskbarAlignment "Center" to "Middle"

mdanish-kh:

~~"Center" value only works at the moment because of #199. If we do some work to tackle #199, the sample would stop working. This PR just changes the value to what the resource currently has logic written against~~

Since "Center" is the value shown in Windows settings, use "Center" string instead of "Middle" in the resource. It doesn't make a different at the moment because of #199, but when we do tackle that issue, the string should be the proper one to use so that the [sample](https://github.com/microsoft/winget-dsc/blob/6e073871037e3719e2a0ce07741892854508d89b/samples/DscResources/Microsoft.Windows.Settings/DefaultWindowsSettings.winget#L19) doesn't break

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-dsc/pull/200)

### DIFF
--- a/resources/Microsoft.Windows.Settings/Microsoft.Windows.Settings.psm1
+++ b/resources/Microsoft.Windows.Settings/Microsoft.Windows.Settings.psm1
@@ -108,11 +108,11 @@ class WindowsSettings {
 
     [string] GetTaskbarAlignment() {
         if (-not(DoesRegistryKeyPropertyExist -Path $global:ExplorerRegistryPath -Name $this.TaskbarAl)) {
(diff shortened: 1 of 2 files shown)


mdanish-kh:
or if "Center" sounds more user-friendly, perhaps we can change the value in the resource . . .

denelon:
Center "should" be the chosen one based on the Windows Settings screenshot.

image

denelon (approved):

PR #219, opened 2025-08-05
Fix publishing to gallery

AmelBawa-msft:
- [ ] I have signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-dsc).

-----
PowerShell@2 doc: https://learn.microsoft.com/en-us/azure/devops/pipelines/tasks/reference/powershell-v2?view=azure-pipelines#inputs

Pipeline error:
```
Publish-Module is not supported on Nano Server and PowerShellCore
```
 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-dsc/pull/219)

### DIFF
--- a/pipelines/azure-pipelines.publish.yml
+++ b/pipelines/azure-pipelines.publish.yml
@@ -133,11 +133,15 @@ extends:
                 *.psd1
 
           steps:
-          - pwsh: |
-              $moduleFolder = "$(System.DefaultWorkingDirectory)/ModuleToPublish/${{ parameters.moduleName }}"
-              Get-ChildItem -Path $moduleFolder -Recurse
-              $moduleFolderPath = (Resolve-Path $moduleFolder).Path
-              Publish-Module -Path $moduleFolderPath -Repository PSGallery -NuGetApiKey $env:api_key -verbose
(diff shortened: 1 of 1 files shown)


pipelines/azure-pipelines.publish.yml
```diff
@@ -133,11 +133,15 @@ extends:
                 *.psd1
 
           steps:
-          - pwsh: |
```
florelis, line 136: You could just replace `pwsh` by `powershell` here. Both are just an alias for `PowerShell@2` but with different values for the `pwsh` input

florelis (commented):
I don't see how this change could fix the error. Using Windows PowerShell instead of PowerShell does not change whether the pipeline is running on Server Nano or not

florelis (approved):

pipelines/azure-pipelines.publish.yml
AmelBawa-msft: Oh didn't know that! Will keep this in mind for the future 😊

PR #232, opened 2025-11-03
Move resource from Microsoft.Windows.Developer to Microsoft.Windows.Settings

AmelBawa-msft:
## Description
- Moved the following resources from Microsoft.Windows.Developer to Microsoft.Windows.Settings
  - WindowsCapability
  - PowerPlanSetting
  - AdvancedNetworkSharingSetting
  - NetConnectionProfile
  - FirewallRule

- [ ] I have signed the [Contributor License Agreement](https://cla.opensource.microsoft.com/microsoft/winget-dsc).

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-dsc/pull/232)

### DIFF
--- a/resources/Microsoft.Windows.Settings/Microsoft.Windows.Settings.psd1
+++ b/resources/Microsoft.Windows.Settings/Microsoft.Windows.Settings.psd1
@@ -12,7 +12,8 @@
         'WindowsCapability',
         'PowerPlanSetting',
         'AdvancedNetworkSharingSetting',
-        'NetConnectionProfile'
+        'NetConnectionProfile',
+        'FirewallRule'
     )
     PrivateData          = @{
         PSData = @{
@@ -22,7 +23,8 @@
                 'PSDscResource_WindowsCapability',
                 'PSDscResource_PowerPlanSetting',
                 'PSDscResource_AdvancedNetworkSharingSetting',
-                'PSDscResource_NetConnectionProfile'
+                'PSDscResource_NetConnectionProfile',
+                'PSDscResource_FirewallRule'
             )
 
             # Prerelease string of this module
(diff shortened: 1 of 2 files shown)


ranm-msft (commented):
This needs a rebase and a reconcile against current `main`, not just a textual conflict resolution. The module and packaging model both changed after this branch was created.

**1. This is currently a copy, not a move.** The diff only touches the two `Microsoft.Windows.Settings` files. Nothing under `resources/Microsoft.Windows.Developer/` is removed, so on `main` today `Microsoft.Windows.Developer.psm1` still declares `WindowsCapability`, `PowerPlanSetting`, `AdvancedNetworkSharingSetting`, `NetConnectionProfile` and `FirewallRule`, and still exports all five from its `.psd1`. After a rebase both modules would ship separate implementations of the same five resource names. DSC can tell them apart by module-qualified type, so coexistence is technically workable, but that is a different outcome from the stated move. Could you either complete the removal from `Microsoft.Windows.Developer` and note the migration path, or say explicitly that both identities stay supported and how the two implementations are kept in sync?

**2. The adapted resource manifests need the same ownership change.** #262 landed in May and added a `*.dsc.manifests.json` per module. `microsoftwindowsdeveloper.dsc.manifests.json` currently advertises all five under `Microsoft.Windows.Developer`, while `microsoftwindowssettings.dsc.manifests.json` advertises only `Microsoft.Windows.Settings/WindowsSettings`. Whichever way point 1 is resolved, those entries and their schemas should be moved or added so the static v3 discovery metadata matches the intended owner.

Worth settling the ownership question before this merges.

PR #249, opened 2026-03-08
NpmDsc: Avoid `Set` false positives with `owner/repo` packages

guillermooo:
The old implementation didn't account for all possible `one/two` package names accepted by `npm install`.

We now limit user input to package names that won't yield false positives in these cases.

NOTES:

The fix is actually restricted to `one/two` type of names; other legal values for a package name may still yield false positives if they reference a code repository. (Untested)

-----

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-dsc/pull/249)

### DIFF
--- a/resources/NpmDsc/NpmDsc.psm1
+++ b/resources/NpmDsc/NpmDsc.psm1
@@ -343,6 +343,14 @@ class NpmPackage {
         $inDesiredState = $this.Test()
         if ($this.Ensure -eq [Ensure]::Present) {
             if (-not $inDesiredState) {
+                # TODO: Handling owner/repo package references pointing to GH repositories requires accounting
+                #       for git errors (missing git, failed authentication, etc.).
+                #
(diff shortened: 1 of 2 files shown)


Gijsreyn:
@guillermooo - thanks for the PR. Can you perhaps include tests in https://github.com/microsoft/winget-dsc/blob/main/tests/NpmDsc/NpmDsc.tests.ps1? The same will be for #250.

guillermooo:
@Gijsreyn Tests added.

Gijsreyn:
> @Gijsreyn Tests added.

I think you need to fix up the spelling checker. @AmelBawa-msft - I think I can't run the Azure Pipelines, mind giving it a spin for Guillermo?

guillermooo:
I'll be sending a new PR with a better fix.

PR #291, opened 2026-06-03
 Change ExtraArgs/Arguments to string[] in GitDsc, NpmDsc, and PythonPip3Dsc

AmelBawa-msft:
## 📖 Description
Internal helpers in GitDsc, NpmDsc, and PythonPip3Dsc previously concatenated DSC property values into a single string and executed it via Invoke-Expression (or, in PythonPip3Dsc, joined with spaces into ProcessStartInfo.Arguments). This PR refactors those helpers to build a [string[]] argument list and invoke the underlying CLI via `& git @args` / `& npm @args` / `ProcessStartInfo.ArgumentList.Add(...)`, so each value is passed as a distinct argv entry with no re-parsing.

## ✅ Checklist

- [ ] Signed the [Contributor License Agreement](https://cla.opensource.microsoft.com)
- [ ] Linked to an issue
- [ ] Updated documentation (if applicable)
- [ ] Updated [Copilot instructions](.github/copilot-instructions.md) (if build, architecture, or conventions changed)

## 📋 Issue Type

- [ ] Bug fix
- [ ] Feature
- [ ] Task

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-dsc/pull/291)

### DIFF
--- a/resources/GitDsc/GitDsc.psm1
+++ b/resources/GitDsc/GitDsc.psm1
@@ -103,7 +103,7 @@ class GitClone {
         if (Test-Path $expectedDirectory) {
             Set-Location -Path $expectedDirectory
             try {
-                $gitRemoteValue = Invoke-GitRemote("get-url $($currentState.RemoteName)")
+                $gitRemoteValue = Invoke-GitRemote -Arguments @('get-url', $currentState.RemoteName)
                 if ($this.HttpsUrl.StartsWith($gitRemoteValue)) {
                     $currentState.Ensure = [Ensure]::Present
                 }
@@ -131,17 +131,18 @@ class GitClone {
 
         Set-Location $this.RootDirectory
 
-        if ($this.FolderName) {
-            $cloneArgs = "$($this.HttpsUrl) $($this.FolderName)"
-        } else {
-            $cloneArgs = $this.HttpsUrl
-        }
-
+        $cloneArgs = [List[string]]::new()
         if ($this.ExtraArgs) {
-            $cloneArgs = "$($this.ExtraArgs) $cloneArgs"
+            foreach ($a in ($this.ExtraArgs -split '\s+' | Where-Object { $_ })) {
+                $cloneArgs.Add($a)
+            }
+        }
+        $cloneArgs.Add($this.HttpsUrl)
+        if ($this.FolderName) {
+            $cloneArgs.Add($this.FolderName)
         }
 
-        Invoke-GitClone($cloneArgs)
+        Invoke-GitClone -Arguments $cloneArgs
     }
 }
 
@@ -206,7 +207,7 @@ class GitRemote {
 
         Set-Location $this.ProjectDirectory
         try {
-            $gitRemoteValue = Invoke-GitRemote("get-url $($this.RemoteName)")
+            $gitRemoteValue = Invoke-GitRemote -Arguments @('get-url', $this.RemoteName)
             $currentState.Ensure = ($gitRemoteValue -like $this.RemoteUrl) ? [Ensure]::Present : [Ensure]::Absent
         } catch {
             $currentState.Ensure = [Ensure]::Absent
@@ -225,13 +226,13 @@ class GitRemote {
 
         if ($this.Ensure -eq [Ensure]::Present) {
             try {
-                Invoke-GitRemote("add $($this.RemoteName) $($this.RemoteUrl)")
+                Invoke-GitRemote -Arguments @('add', $this.RemoteName, $this.RemoteUrl)
             } catch {
                 throw 'Failed to add remote repository.'
             }
         } else {
             try {
-                Invoke-GitRemote("remove $($this.RemoteName)")
+                Invoke-GitRemote -Arguments @('remove', $this.RemoteName)
             } catch {
                 throw 'Failed to remove remote repository.'
             }
@@ -305,8 +306,8 @@ class GitConfigUserName {
             }
         }
 
-        $configArgs = ConstructGitConfigUserArguments -Arguments 'user.name' -ConfigLocation $this.ConfigLocation
-        $result = Invoke-GitConfig($configArgs)
+        $configArgs = ConstructGitConfigUserArguments -Arguments @('user.name') -ConfigLocation $this.ConfigLocation
+        $result = Invoke-GitConfig -Arguments $configArgs
         $currentState.Ensure = ($currentState.UserName -eq $result) ? [Ensure]::Present : [Ensure]::Absent
         return $currentState
     }
@@ -326,12 +327,12 @@ class GitConfigUserName {
         }
 
         if ($this.Ensure -eq [Ensure]::Present) {
-            $configArgs = ConstructGitConfigUserArguments -Arguments "user.name '$($this.UserName)'" -ConfigLocation $this.ConfigLocation
+            $configArgs = ConstructGitConfigUserArguments -Arguments @('user.name', $this.UserName) -ConfigLocation $this.ConfigLocation
         } else {
-            $configArgs = ConstructGitConfigUserArguments -Arguments '--unset user.name' -ConfigLocation $this.ConfigLocation
+            $configArgs = ConstructGitConfigUserArguments -Arguments @('--unset', 'user.name') -ConfigLocation $this.ConfigLocation
         }
 
-        Invoke-GitConfig($configArgs)
+        Invoke-GitConfig -Arguments $configArgs
     }
 }
 
@@ -401,8 +402,8 @@ class GitConfigUserEmail {
             }
         }
 
-        $configArgs = ConstructGitConfigUserArguments -Arguments 'user.email' -ConfigLocation $this.ConfigLocation
-        $result = Invoke-GitConfig($configArgs)
+        $configArgs = ConstructGitConfigUserArguments -Arguments @('user.email') -ConfigLocation $this.ConfigLocation
+        $result = Invoke-GitConfig -Arguments $configArgs
         $currentState.Ensure = ($currentState.UserEmail -eq $result) ? [Ensure]::Present : [Ensure]::Absent
         return $currentState
     }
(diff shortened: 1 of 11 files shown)


resources/GitDsc/GitDsc.psm1
```diff
@@ -131,17 +131,18 @@ class GitClone {
...
         Set-Location $this.RootDirectory
 
-        if ($this.FolderName) {
-            $cloneArgs = "$($this.HttpsUrl) $($this.FolderName)"
-        } else {
-            $cloneArgs = $this.HttpsUrl
-        }
-
+        $cloneArgs = [List[string]]::new()
         if ($this.ExtraArgs) {
```
AmelBawa-msft, line 145: Instead of splitting a string by spaces, we could introduce a breaking change to the ExtraArgs DSC property so that it accepts a string array rather than a single string. Thoughts?

resources/PythonPip3Dsc/PythonPip3Dsc.psm1
```diff
@@ -25,7 +24,9 @@ function Invoke-Process {
         $pinfo.UseShellExecute = $false
         $pinfo.WindowStyle = 'Hidden'
         $pinfo.CreateNoWindow = $true
-        $pinfo.Arguments = $ArgumentList
+        foreach ($arg in @($ArgumentList | Where-Object { $null -ne $_ })) {
+            [void]$pinfo.ArgumentList.Add([string]$arg)
+        }
```
florelis, line 29:
```suggestion
        $pinfo.ArgumentList = $ArgumentList | Where-Object { $null -ne $_ }
```

resources/PythonPip3Dsc/PythonPip3Dsc.psm1
```diff
@@ -171,8 +172,10 @@ function Invoke-Pip3Install {
         $command.Add('--dry-run')
     }
 
-    $command.Add($Arguments)
-    Write-Verbose -Message "Executing 'pip' install with command: $command"
+    foreach ($a in ($Arguments -split '\s+' | Where-Object { $_ })) {
+        $command.Add($a)
+    }
```
florelis:
```suggestion
    $command += ($Arguments -split '\s+') | Where-Object { $_ }
```

resources/PythonPip3Dsc/PythonPip3Dsc.psm1
```diff
@@ -193,7 +196,9 @@ function Invoke-Pip3Uninstall {
     $command = [List[string]]::new()
     $command.Add('uninstall')
     $command.Add((Get-PackageNameWithVersion -PackageName $PackageName -Version $Version))
-    $command.Add($Arguments)
+    foreach ($a in ($Arguments -split '\s+' | Where-Object { $_ })) {
+        $command.Add($a)
+    }
```
florelis:
```suggestion
    $command += ($Arguments -split '\s+') | Where-Object { $_ }
```

resources/PythonPip3Dsc/PythonPip3Dsc.psm1
```diff
@@ -242,17 +247,15 @@ function GetPip3CurrentState {
...
     if ($global:usePip3Exe) {
-        $command = "& '$global:pip3ExePath' " + $Arguments
+        $proc = Invoke-Process -FilePath $global:pip3ExePath -ArgumentList $arguments
     } else {
-        $command = '& pip3 ' + $Arguments
+        $proc = Invoke-Process -FilePath 'pip3' -ArgumentList $arguments
     }
 
-    $res = Invoke-Expression -Command $command | ConvertFrom-Json
+    $res = ($proc.StdOut -join "`n") | ConvertFrom-Json
```
florelis, line 255: I think you could avoid writing the `Invoke-Process` twice, although I'm not sure this is better.

```suggestion
    $local:pip3ExePath = 'pip3'
    if ($global:usePip3Exe) {
        $local:pip3ExePath = $global:pip3ExePath
    }
    
    $proc = Invoke-Process -FilePath $global:pip3ExePath -ArgumentList $arguments
    $res = ($proc.StdOut -join "`n") | ConvertFrom-Json
```

resources/GitDsc/GitDsc.psm1
florelis: What does this look in practice?

The syntax is like yaml, right? What does it look like now and how would it change?

`ExtraArgs: --some arg`
or
`ExtraArgs: "--some arg"`
or
```
ExtraArgs:
  - --some
  - args
```
?

resources/GitDsc/GitDsc.psm1
```diff
@@ -131,17 +131,18 @@ class GitClone {
...
-            $cloneArgs = "$($this.ExtraArgs) $cloneArgs"
+            foreach ($a in ($this.ExtraArgs -split '\s+' | Where-Object { $_ })) {
+                $cloneArgs.Add($a)
+            }
```
florelis:
```suggestion
            $cloneArgs += ($this.ExtraArgs -split '\s+') | Where-Object { $_ }
```

resources/NpmDsc/NpmDsc.psm1
```diff
@@ -104,9 +110,11 @@ function Uninstall-NpmPackage {
         $command.Add('-g')
     }
 
-    $command.Add($Arguments)
+    foreach ($a in ($Arguments -split '\s+' | Where-Object { $_ })) {
+        $command.Add($a)
+    }
```
florelis:
```suggestion
    $command += ($Arguments -split '\s+') | Where-Object { $_ }
```

resources/GitDsc/GitDsc.psm1
```diff
@@ -131,17 +131,18 @@ class GitClone {
...
-        if ($this.FolderName) {
-            $cloneArgs = "$($this.HttpsUrl) $($this.FolderName)"
-        } else {
-            $cloneArgs = $this.HttpsUrl
-        }
-
+        $cloneArgs = [List[string]]::new()
         if ($this.ExtraArgs) {
-            $cloneArgs = "$($this.ExtraArgs) $cloneArgs"
+            foreach ($a in ($this.ExtraArgs -split '\s+' | Where-Object { $_ })) {
```
florelis: What values do we expect in the `ExtraArgs`?

If someone has something like `ExtraArgs: 'some "value with" spaces'` then it should be split as `some`, `"value with"`, `spaces`, but this would split as `some`, `"value`, `with"`, `spaces`. That would be a breaking change from just passing it to the command as a string

florelis (commented):
Made some suggestions, but I'm not sure if they're correct.

resources/GitDsc/GitDsc.psm1
AmelBawa-msft: Exactly the case that pushed me to the array approach 😊 So now with string array we don't have to parse the tricky quotes and spaces and maintain it. Each element is passed as a discrete argument, so a value like `user.name=Some Name` works as expected

resources/GitDsc/GitDsc.psm1
AmelBawa-msft: Before:
 ```yaml
  ExtraArgs: --depth 1 --no-checkout
 ```

 After:
``` yaml
  ExtraArgs:
    - --depth
    - "1"
    - --no-checkout
```

resources/GitDsc/GitDsc.psm1
AmelBawa-msft: Resolving since `ExtraArgs` is now an array instead

resources/NpmDsc/NpmDsc.psm1
AmelBawa-msft: Resolving since `Arguments` is now an array instead

resources/PythonPip3Dsc/PythonPip3Dsc.psm1
AmelBawa-msft: Resolving since ExtraArgs is now an array instead

resources/PythonPip3Dsc/PythonPip3Dsc.psm1
AmelBawa-msft: Resolving since Arguments is now an array instead

resources/PythonPip3Dsc/PythonPip3Dsc.psm1
AmelBawa-msft: I can't assign directly, `ProcessStartInfo.ArgumentList` is a read-only `Collection<string>`

florelis: I see. That's unfortunate; it feels silly to have to write a foreach each time

resources/NpmDsc/NpmDsc.psm1
```diff
@@ -15,13 +15,15 @@ function Assert-Npm {
...
     )
-    $value = Invoke-Expression -Command "npm $Command"
+
+    $argList = @($Command | Where-Object { -not [string]::IsNullOrEmpty($_) })
+    $value = & npm @argList
 
     if ($LASTEXITCODE -ne 0) {
         $errors = Get-NpmErrorMessages -LogPath (GetNpmPath)
-        throw "Command 'npm $($Command.Trim())' failed: $($errors -join '; ')"
+        throw "Command 'npm $($argList -join ' ')' failed: $($errors -join '; ')"
```
florelis, line 26: nit: we could print this with the quotes around individual arguments for clarity in case any of them has spaces. `Join-String -DoubleQuote -Separator ' '`

resources/GitDsc/GitDsc.psm1
```diff
@@ -56,6 +57,15 @@ enum ConfigLocation {
         }
 
         This example clones the winget-dsc repository into C:\repos.
+
+    .EXAMPLE
+        Invoke-DscResource -ModuleName GitDsc -Name GitClone -Method Set -Property @{
+            HttpsUrl      = 'https://github.com/microsoft/winget-dsc'
+            RootDirectory = 'C:\repos'
+            ExtraArgs     = @('--depth', '1')
```
florelis, line 65: I believe you won't need the `@()` here. You could just do `ExtraArgs = '--depth', '1'`, which may be easier

florelis (approved):

PR #309, opened 2026-08-19
Pin GitHub Actions to full-length commit SHAs

danfiedler-msft:
## Summary

This PR pins GitHub Actions to full-length commit SHAs for improved security and reproducibility and adds a 7 day cooldown to Dependabot configuration for GitHub Actions. This work is described in more detail at https://aka.ms/action-pinning.

## Why?

Pinning actions to commit SHAs prevents supply-chain attacks where a tag could be moved to point to malicious code. This is a recommended security best practice per the [GitHub Actions security hardening guide](https://docs.github.com/en/actions/security-for-github-actions/security-guides/security-hardening-for-github-actions#using-third-party-actions).

This change mitigates the risk of tag retargeting to malicious code as seen in incidents like the [tj-actions/changed-files action compromise](https://www.stepsecurity.io/blog/harden-runner-detection-tj-actions-changed-files-action-is-compromised) or [codfish/semantic-release-action compromise](https://www.stepsecurity.io/blog/supply-chain-compromise-codfish-semantic-release-action) and improves the integrity and reproducibility of the CI/CD pipeline.

## What changed?

**Action pinning:** Third-party action references in `.github/workflows/` that used mutable tag-based references (e.g., `actions/checkout@v4`) have been updated to full-length commit SHAs with a version comment (e.g., `actions/checkout@<sha> # v4`) using the [pinact](https://github.com/suzuki-shunsuke/pinact) tool. References that were already pinned to a SHA, or that used immutable release tags, were left unchanged.

**Dependabot configuration:** `.github/dependabot.yml` has been updated to ensure a `github-actions` package-ecosystem section is present with a `cooldown` configuration (`default-days: 7`). If the file did not exist, it was created. If a `github-actions` section already existed, only the `cooldown` block was added or its `default-days` value was increased to 7 if it was lower. The 7-day cooldown provides a window for the community to detect and report compromised releases before they are automatically proposed as updates, reducing exposure to supply-chain attacks via newly published malicious versions.

## Is this safe to merge?

Yes. The pinned SHAs correspond to the same commits that the existing tags pointed to. No behavioral changes in action execution are introduced. You can verify the pinned SHA value using the GitHub REST API (e.g., the commit hash for `actions/checkout@v7` can be found in the `sha` property in the JSON response for `GET https://api.github.com/repos/actions/checkout/commits/v7`).

## Additional Information

For more information, please see https://aka.ms/action-pinning

 ###### Microsoft Reviewers: [Open in CodeFlow](https://microsoft.github.io/open-pr/?codeflow=https://github.com/microsoft/winget-dsc/pull/309)


ranm-msft (commented):
Thanks for this - I verified the mechanical substance rather than skimming it.

**Every pinned SHA is correct.** For each `uses:` I resolved the pinned commit through the GitHub API (dereferencing annotated tag objects to their commits) and confirmed it matches the tag its comment claims:

- `actions/checkout@11d5960` = v4.4.0
- `tj-actions/changed-files@48d8f15` = v45.0.2
- `DavidAnson/markdownlint-cli2-action@db43aef` = v17.0.0
- `check-spelling/check-spelling@cfb6f7e` = v0.0.26

I also compared each pinned SHA against what the *original* floating ref resolves to today - all identical. So this is behavior-preserving: no silent downgrade, no version drift. I checked coverage too, and `duplicate-surfacing.lock.yml` was already fully pinned (53/53), so the repo is complete after this.

One thing blocking merge:

**`Check Spelling` is failing on this PR**, and the newly added file is the cause:

```
.github/dependabot.yml:10  `cooldown` is not a recognized word (unrecognized-spelling)
```

The config itself is fine - `cooldown` / `default-days` is valid, documented Dependabot syntax. The repo's dictionary just doesn't know the word. Adding `cooldown` to `.github/actions/spelling/expect/generic_terms.txt` should turn it green.

Since you're rolling this out across many repos, this likely reproduces anywhere check-spelling is enabled - is it worth folding into the campaign template so each owner doesn't hit it separately?

Happy to approve as soon as the check is green.

ranm-msft (approved):
Approved.

I re-verified all four SHA/version mappings at the new head, and they all still match what their comments claim:

- `actions/checkout@11d5960` = v4.4.0
- `tj-actions/changed-files@48d8f15` = v45.0.2
- `DavidAnson/markdownlint-cli2-action@db43aef` = v17.0.0
- `check-spelling/check-spelling@cfb6f7e` = v0.0.26

The only delta since my earlier review is @denelon's follow-up commit, and it is a single line adding `cooldown` to `.github/actions/spelling/expect/generic_terms.txt` - exactly the fix that was needed. `Check Spelling` is green now, alongside `Report (PR)` and `license/cla`.

That clears the one thing I had flagged, so this is good to go from my side. Thanks for the clean, well-documented change, and for the verification steps in the description - they made this quick to confirm rather than take on faith.

